PR #29, opened 2024-04-25
Tanya/test and gitleaks fix

Tanya-atatakai:
## Types of changes


apps/tokenomics/.eslintrc.json
```diff
@@ -26,12 +26,6 @@
     {
       "files": ["*.js", "*.jsx"],
       "rules": {}
-    },
```
Tanya-atatakai, line 29: this comes from the `"../../.eslintrc.json"` anyway, so I removed from here

.github/workflows/tokenomics.yml
```diff
@@ -21,27 +21,12 @@ jobs:
       - uses: actions/checkout@v4
         with:
           fetch-depth: 0
-      - uses: actions/setup-node@v3
+      - uses: actions/setup-node@v4
         with:
-          node-version: 18
+          node-version: '20.x'
```
Tanya-atatakai, line 26: updated node version here as well as we decided to move tokenomics to 20 (though we didn't do that in Vercel, neither for current tokenomics, nor for monorepo, do you want me to do that? @mohandast52 )

Tanya-atatakai: I'd start with the monorepo as it's not in prod :D

PR #33, opened 2024-06-11
Refactor queries to ensure order direction in ListAgents, ListComponents, and ListServices

truemiller:
## Proposed changes

Update order direction to `desc` for Components, Agents, Services.

## Types of changes


oaksprout (approved):

truemiller:
> down arrow maybe

good catch, I hadn't noticed this on other chains. will see what's causing this.
currently ticket is just to sort by 'desc', happy to make this dynamic so user can click on the header, thoughts here @oaksprout?

oaksprout (approved):
Re-tested. Looks good.

PR #37, opened 2024-06-17
Show alert when on mobile

Tanya-atatakai:
## Proposed changes

1) Moves useScreen hook from `@autonolas/frontend-library` to shared libs
2) Displays alert on mobile screens

image

## Fixes

`isTablet` didn't work correctly in `@autonolas/frontend-library` because it checked `!sm` and `!xs` when they both were true (see at the pic what values are true when on "lg" screen)

Screenshot 2024-06-17 at 11 47 11

## Types of changes

### DIFF
--- a/apps/govern/components/Layout/index.tsx
+++ b/apps/govern/components/Layout/index.tsx
@@ -37,7 +37,7 @@ export const Layout: FC<LayoutProps> = ({ children }) => {
         {isMobile && (
           <Alert
             className="mt-16"
-            message="This page is not optimized for mobile. For best experience, please come back on desktop."
+            message="This page is not optimized for mobile. For the best experience, please come back on desktop."
             showIcon
             type="warning"
           />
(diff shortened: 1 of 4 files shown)


libs/ui-theme/src/lib/hooks/useScreen.ts
```diff
@@ -0,0 +1,12 @@
+import { Grid } from 'antd';
+
+const { useBreakpoint } = Grid;
+
+export const useScreen = () => {
```
Tanya-atatakai, line 5: please lmk if you think the `libs/ui-theme` is not the correct library for this hook

apps/govern/components/Layout/index.tsx
```diff
@@ -30,6 +34,15 @@ export const Layout: FC<LayoutProps> = ({ children }) => {
       </OlasHeader>
 
       <Content className="site-layout">
+        {isMobile && (
+          <Alert
+            className="mt-16"
+            message="This page is not optimized for mobile. For best experience, please come back on desktop."
```
truemiller, line 40:
```suggestion
            message="This page is not optimized for mobile. For the best experience, please come back on desktop."
```

truemiller (approved):

oaksprout (approved):

PR #40, opened 2024-06-18
feat: Request all user’s contracts on selected chain

mohandast52:
## Proposed changes

- List of my staking contracts

S2

https://github.com/valory-xyz/autonolas-frontend-mono/assets/22061815/d551b87b-9b69-485b-b903-21084ba7c956

## Types of changes

### DIFF
--- a/apps/launch/common-util/constants/rpcs.ts
+++ b/apps/launch/common-util/constants/rpcs.ts
@@ -1,7 +1,5 @@
 import { RPC_URLS } from 'libs/util-constants/src';
 
-export type SupportedChain = 1 | 100 | 137; // TODO: make it dynamic
-
 export const LAUNCH_RPC_URLS: Record<number, string> = {
   1: RPC_URLS[1],
   100: RPC_URLS[100],
--- a/apps/launch/common-util/constants/stakingContract.ts
+++ b/apps/launch/common-util/constants/stakingContract.ts
@@ -103,11 +103,11 @@ export const isSupportedChainId = (chainId: number): chainId is ChainId => {
   return chainId in IMPLEMENTATION_ADDRESSES;
 };
 
-type BlockNumber = {
+type BlockNumbers = {
   [key in ChainId]: number;
 };
 
-export const blockNumbers: BlockNumber = {
+export const blockNumbers: BlockNumbers = {
   [mainnet.id]: 20009952,
   [optimism.id]: 0,
   [gnosis.id]: 34225443,
--- a/apps/launch/hooks/useGetMyStakingContracts.ts
+++ b/apps/launch/hooks/useGetMyStakingContracts.ts
@@ -77,9 +77,9 @@ const useGetMyStakingContractsMetadata = (addresses: Address[]) => {
             if (response.status === 'fulfilled' && response.value) {
               results.push(response.value);
             } else {
-              // to maintain the order of the addresses, push null
+              // To maintain the order of the addresses, push null.
               // Eg: if the metadata for the 2nd address is not available or failed to fetch,
-              //  push null at 2nd index
+              // push null at 2nd index
               results.push(null);
             }
           }
@@ -133,7 +133,7 @@ const useGetInstanceAddresses = () => {
           toBlock: 'latest',
         });
 
-        // log.args[1] is the instance address
+        // log.args[1] = instance address
         const addresses = eventLogs.map((log) => log.args[1] as Address);
         setInstanceAddresses(addresses);
       } catch (e) {
@@ -179,7 +179,7 @@ export const useGetMyStakingContracts = () => {
         };
       })
       // filter out null values (ie. contracts without metadata - failed to fetch or not available)
-      .filter((c): c is MyStakingContract => c !== null);
+      .filter((contract): contract is MyStakingContract => contract !== null);
 
     dispatch(setMyStakingContracts(myStakingContractsList));
   }, [


apps/launch/hooks/useGetMyStakingContracts.ts
```diff
@@ -0,0 +1,168 @@
...
+  }));
+
+  const { data } = useReadContracts({
+    contracts,
+    query: {
+      enabled: addresses.length > 0,
+      select: async (list) => {
+        const metadataHashList = list.map(({ result }) => result as string);
+        const metadataPromise = await Promise.allSettled(
+          metadataHashList.map(async (hash) => await getMetadata(hash)),
```
truemiller, line 77: not sure this is required

```suggestion
          metadataHashList.map(async (hash) => getMetadata(hash)),
```

apps/launch/hooks/useGetMyStakingContracts.ts
```diff
@@ -0,0 +1,168 @@
...
+    args: [],
+  }));
+
+  const { data } = useReadContracts({
+    contracts,
+    query: {
+      enabled: addresses.length > 0,
+      select: async (list) => {
+        const metadataHashList = list.map(({ result }) => result as string);
+        const metadataPromise = await Promise.allSettled(
```
truemiller, line 76: no longer a promise if resolved with await

```suggestion
        const metadata = await Promise.allSettled(
```

apps/autonolas-registry/components/Layout/styles.jsx
```diff
@@ -29,6 +30,20 @@ export const CustomLayout = styled(Layout)`
...
   }
 
+  /* table */
+  .ant-table {
+    .ant-table-tbody > tr {
+      > td {
+        padding: 8px 16px;
+        .ant-btn {
+          &:first-child {
+            padding-left: 0;
```
Tanya-atatakai, line 40: struggled with in Govern and wanted to suggest moving to shared libs but realise we have it here. do you have any idea why it doesn't work from there?

apps/launch/common-util/constants/rpcs.ts
```diff
@@ -1,5 +1,7 @@
 import { RPC_URLS } from 'libs/util-constants/src';
 
+export type SupportedChain = 1 | 100 | 137; // TODO: make it dynamic
```
Tanya-atatakai, line 3: feel free to pull main, should be there https://github.com/valory-xyz/autonolas-frontend-mono/blob/main/apps/launch/common-util/constants/stakingContract.ts#L37

apps/launch/components/MyStakingContracts/NominateContract.tsx
```diff
@@ -0,0 +1,54 @@
...
+const PurpleDot = styled.div`
+  width: 8px;
+  height: 8px;
+  border-radius: 50%;
+  background-color: ${COLOR.PRIMARY};
+  display: inline-block;
+  margin-right: 8px;
+`;
+
+// TODO: Tanya to check
```
Tanya-atatakai, line 33: thanks!

apps/launch/components/MyStakingContracts/index.tsx
```diff
@@ -21,66 +22,96 @@ const ICON_STYLE = { fontSize: '56px', color: '#A3AEBB' };
...
 
       <Card className="flex-none mt-24">
-        <Title level={3} className="m-0">
-          My staking contracts
-          {isAccountConnected && networkId && <>&nbsp;on {CHAIN_NAMES[networkId]}</>}
-        </Title>
+        <Flex justify="space-between">
+          <Title level={4} className="m-0">
+            My staking contracts
+            {isAccountConnected && networkId && <>&nbsp;on {CHAIN_NAMES[networkId]}</>}
```
Tanya-atatakai, line 109: can just  ```{isAccountConnected && networkId && ` on ${CHAIN_NAMES[networkId]}`}``` ?

apps/launch/components/MyStakingContracts/index.tsx
```diff
@@ -21,66 +22,96 @@ const ICON_STYLE = { fontSize: '56px', color: '#A3AEBB' };
...
+  </Paragraph>
+);
+
+const CreateContractButton = () => {
   const router = useRouter();
   const { networkName } = useAppSelector((state) => state.network);
-  console.log('networkName', `${networkName}/${URL.myStakingContracts}/create`);
 
   return (
+    <Button
```
Tanya-atatakai, line 61: I think we can just use Link with Button inside, so no need to use router hook and do push, wdyt?
```
<Link href={`/${networkName}/${URL.myStakingContracts}/create`} passHref>
    <Button>Create staking contract</Button>
</Link>
```

apps/launch/hooks/useGetMyStakingContracts.ts
```diff
@@ -0,0 +1,175 @@
...
+  const { data: nominees } = useReadContract({
+    address: (VOTE_WEIGHTING.addresses as Record<number, Address>)[mainnet.id],
+    abi: VOTE_WEIGHTING.abi,
+    chainId: mainnet.id,
+    functionName: 'getAllNominees',
+    args: [],
+    query: {
+      select: (data: unknown) => {
+        const list = data as { account: Address; chainId: bigint }[];
+        return list.filter((nominee) => nominee.chainId === BigInt(mainnet.id));
```
Tanya-atatakai, line 50: I don't think we need to filter only those on mainnet, in this case we won't see any nominated contracts e.g. on gnosis. Vote Weighting contracts live only on mainnet, but contains all nominees from different chains with chainId specified, so we can check if any contracts on different chains are nominated

apps/launch/hooks/useGetMyStakingContracts.ts
```diff
@@ -0,0 +1,175 @@
...
+};
+
+const useGetAllNominees = () => {
+  // it is only available on mainnet, hence hardcoded.
+  const { data: nominees } = useReadContract({
+    address: (VOTE_WEIGHTING.addresses as Record<number, Address>)[mainnet.id],
+    abi: VOTE_WEIGHTING.abi,
+    chainId: mainnet.id,
+    functionName: 'getAllNominees',
+    args: [],
```
Tanya-atatakai, line 46: no need to pass args if the list is empty, can remove?

apps/launch/hooks/useGetMyStakingContracts.ts
```diff
@@ -0,0 +1,175 @@
...
+const useGetMyStakingContractsMetadata = (addresses: Address[]) => {
+  const { chainId } = useAccount();
+  const [metadata, setMetadata] = useState<Metadata[]>([]);
+
+  const contracts = addresses.map((address) => ({
+    address,
+    abi: STAKING_TOKEN.abi as Abi,
+    chainId,
+    functionName: 'metadataHash',
+    args: [],
```
Tanya-atatakai, line 67: tiny: can also remove

apps/launch/hooks/useGetMyStakingContracts.ts
```diff
@@ -0,0 +1,175 @@
...
+    getBytes32FromAddress(address),
+  );
+
+  useEffect(() => {
+    if (!myStakingContractsMetadata) return;
+    if (myStakingContractsMetadata.length === 0) return;
+    if (!nominees) return;
+
+    const myStakingContracts: MyStakingContract[] = myStakingContractsMetadata.map((metadata) => ({
+      id: kebabCase(metadata.name),
```
Tanya-atatakai, line 165: ideally we need to have here instance address - will be needed for nomination and displaying contract page

apps/launch/hooks/useGetMyStakingContracts.ts
```diff
@@ -0,0 +1,175 @@
...
+        const metadataHashList = list.map(({ result }) => result as string);
+        try {
+          const metadataList = await Promise.allSettled<Metadata>(
+            metadataHashList.map((hash) => getMetadata(hash)),
+          );
+
+          const results: Metadata[] = [];
+          for (const response of metadataList) {
+            if (response.status === 'fulfilled' && response.value) {
+              results.push(response.value);
```
Tanya-atatakai, line 84: Since we use allSettled, I think it makes sense to track which instance address this metadata corresponds to. Without that, we'll be unable to identify which address this metadata belongs to

apps/launch/hooks/useGetMyStakingContracts.ts
```diff
@@ -0,0 +1,175 @@
...
+      }
+    })();
+  }, [data, addresses]);
+
+  return metadata;
+};
+
+const useGetInstanceAddresses = () => {
+  const client = usePublicClient();
+  const { chainId } = useAccount();
```
Tanya-atatakai, line 116: I think we should use network from the route instead, because user may not change their network in wallet but we still should show what's selected from the dropdown

apps/launch/hooks/useGetMyStakingContracts.ts
```diff
@@ -0,0 +1,175 @@
...
+      if (!currentChainId) return;
+      if (!client) return;
+
+      try {
+        const eventLogs = await client.getLogs({
+          address: STAKING_FACTORY.addresses[`${currentChainId}`] as Address,
+          event: parseAbiItem(
+            'event InstanceCreated(address indexed, address indexed, address indexed)',
+          ),
+          fromBlock: BigInt(0),
```
Tanya-atatakai, line 133: can we hardcode the block to the contract deployed one? can change when real contracts will be deployed.

didn't find the exact blocks, but the closes are:
Polygon: 57751430,
Gnosis: 34225443
Ethereum: 20009952

apps/launch/hooks/useGetMyStakingContracts.ts
```diff
@@ -0,0 +1,175 @@
...
+  const nominees = useGetAllNominees();
+
+  const instanceAddressesInBytes32 = instanceAddresses.map((address) =>
+    getBytes32FromAddress(address),
+  );
+
+  useEffect(() => {
+    if (!myStakingContractsMetadata) return;
+    if (myStakingContractsMetadata.length === 0) return;
+    if (!nominees) return;
```
Tanya-atatakai, line 162: can we also check that the myStakingContracts is empty, this hook will be triggered again on each page change and without this check the data will be recalculated and dispatched again

apps/launch/hooks/useGetMyStakingContracts.ts
Tanya-atatakai: 1) in my virtual test nets
2) I deployed them myself, for testing. They are not deployed yet in real blockchain (neither factories, nor staking tokens, vote weighting, etc.)

apps/launch/components/MyStakingContracts/index.tsx
Tanya-atatakai: 1) less code
2) no need to use route hook (doesn't change much though, but could in case we have e.g. filtering that changes the url)
3) on each component render the onClick function won't be recreated (cuz there will be no onClick passed)

apps/launch/common-util/constants/rpcs.ts
Tanya-atatakai: ah I see. it's expected b/c I only deployed contracts on those three chains, but on prod we're supposed to have addresses for all 7 chains. hmm maybe we can add some dummy addresses for now for each remaining chain?

apps/launch/common-util/constants/stakingContract.ts
```diff
@@ -102,3 +102,17 @@ export const STAKING_TOKEN_ADDRESSES: Addresses = {
...
   return chainId in IMPLEMENTATION_ADDRESSES;
 };
+
+type BlockNumber = {
+  [key in ChainId]: number;
+};
+
+export const blockNumbers: BlockNumber = {
+  [mainnet.id]: 20009952,
+  [optimism.id]: 0,
```
truemiller, line 112: 0 is a valid block height, might be worth changing to null if it represents contract not being deployed or something similar

truemiller (approved):

Tanya-atatakai (approved):

oaksprout (approved):

PR #44, opened 2024-06-24
Fix outdated hash on details page

Tanya-atatakai:
## Fixes
Context: We display the metadata hash on listing pages and detail pages, but we get them from different sources: for listing pages - from the registry subgraph, for detail pages - from the contract state. One user after updating their metadata hash found that it wasn't updated on the detail page.

Fix: check if there are updated hashes and get the last one, otherwise get the initial one.
https://github.com/valory-xyz/autonolas-registries/blob/main/contracts/UnitRegistry.sol#L171

## Types of changes

### DIFF
--- a/apps/autonolas-registry/components/ListComponents/utils.jsx
+++ b/apps/autonolas-registry/components/ListComponents/utils.jsx
@@ -1,8 +1,7 @@
-import {
-  getMechMinterContract,
-  getComponentContract,
-} from 'common-util/Contracts';
+import { GATEWAY_URL, HASH_PREFIX } from 'libs/util-constants/src';
+
 import { getListByAccount } from 'common-util/ContractUtils/myList';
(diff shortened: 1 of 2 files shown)


kupermind (approved):

truemiller (approved):

Tanya-atatakai:
@mohandast52 fixed tests, but the fix is weird (I just updated the import), but the error is also weird. the problem is probably in our jest setup, didn't find quicker solution.

> how did you find the discrepancy
1) I checked the correct hash in the tx logs provided by the user https://etherscan.io/tx/0xe05a006bae5b7ed0e0e107af95a9f4195fe8fcdc23edf0230ea303666a0bd25f
2) checked what methods the contract provides + the code and the comments + tested some examples https://etherscan.io/address/0x15bd56669F57192a97dF41A2aa8f4403e9491776#readContract

PR #49, opened 2024-06-26
Govern footer

Tanya-atatakai:
## Proposed changes

1) Created ui-components library where we can store shared components
2) Added footer to it, fixed tests setup and a little improved tests. Ideally we should have a separate github workflow for libs tests, gonna do that as part of tests task for govern
3) Restyled Footer, now it has all icons inside, we only need to provide github url
4) Added contracts list
5) Added disclaimer page

Footer:
image

Tests:
image

## Types of changes


libs/ui-components/src/lib/Footer/index.tsx
Tanya-atatakai, line 1: I noticed that we have Footer and some other component in `libs/feature-service-status-info`. Didn't find where the lib is used, and by its name it's hard to figure out that it contains a footer component. @mohandast52 Do you remember what was the purpose of this lib? can we remove it maybe?

jest.preset.js
```diff
@@ -1,3 +1,11 @@
...
-
-module.exports = { ...nxPreset };
+module.exports = {
+  ...nxPreset,
+  testMatch: ['**/+(*.)+(spec|test).+(ts|js)?(x)'],
+  transform: {
+    '^.+\\.(ts|js|html)$': 'ts-jest',
+  },
+  resolver: '@nx/jest/plugins/resolver',
+  moduleFileExtensions: ['ts', 'js', 'html'],
```
truemiller, line 9: don't know much about this, but do we need to include jsx and tsx?

libs/ui-components/src/lib/Footer/Socials.tsx
```diff
@@ -0,0 +1,56 @@
+import { Button, Flex } from 'antd';
+
+const WebSvg = () => (
```
truemiller, line 3: please export these svgs to .svg files

truemiller (approved):
minor comments, approved 👍

jest.preset.js
Tanya-atatakai: I guess yes, some of the old apps are still written in js... e.g. registry

libs/ui-components/src/lib/Footer/Socials.tsx
Tanya-atatakai: I wanted to do it this way first, but there are problems with making them visible in apps... looks like we need to add a particular setup to each app's project.json https://nx.dev/recipes/tips-n-tricks/include-assets-in-build (easy to forget with many apps) or create a custom builder that will use this assets config, or even add some script that will copy all the assets https://stackoverflow.com/questions/73343438/nx-shared-asset-library 
all those I found more complicated than just having svgs as component, at least for now

Tanya-atatakai: there's a way to have a global setup actually (so no need to copy to each app), but this will copy all the assets from the lib to the app regardless of whether it uses new footer or not 

anyway, lmkyt @truemiller @mohandast52

apps/govern/common-util/constants/urls.ts
```diff
@@ -1 +1,3 @@
 export const GET_VEOLAS_URL = 'https://member.olas.network/';
+export const REPO_URL =
+  'https://github.com/valory-xyz/autonolas-frontend-mono/tree/main/apps/govern';
```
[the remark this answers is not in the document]
truemiller: the file structure here is pretty clear, not sure what this would achieve

libs/ui-components/src/lib/Footer/Socials.tsx
truemiller: can we not just add them to a shared lib called "images"?
i don't have too much nx experience

apps/govern/common-util/constants/urls.ts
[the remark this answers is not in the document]
Tanya-atatakai: agree (with Josh). actually, "utils" sounds to me more like helper functions, I think even hooks are not included - should be separated. but this is just what I'm used to

libs/ui-components/jest.config.ts
```diff
@@ -0,0 +1,11 @@
+/* eslint-disable */
+export default {
+  displayName: 'ui-components',
```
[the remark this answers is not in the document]
Tanya-atatakai: it's gonna be as generic as "util-constants" or "util-functions" - for all shared components. I don't think libs should be very specific, it's up to us right? in nx examples they just combine all those (functions, components, styles) to one folder called "shared" + have feature(or page) specific folders https://github.com/nrwl/nx-examples/tree/master/libs

libs/ui-components/src/lib/Footer/Socials.tsx
Tanya-atatakai: we can, but we'll have the same problem with configs - the images need to be added to each app somehow, so they are accessible from it. the only way to do it is to add a special config which will copy all of them to each app images folder

apps/govern/common-util/constants/urls.ts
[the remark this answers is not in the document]
Tanya-atatakai: good point about the app urls (but not the repo urls maybe, at least I don't see where except in its own app it can be used)
but we can keep them in the `libs/util-constants/src/lib/urls` or `libs/util-constants/src/lib/appUrls`? wdyt

dagacha (approved):

PR #64, opened 2024-07-16
(govern) fix: remove debug code

Tanya-atatakai:
## Fixes

1) fixes removed nominee check - with previous code we always displayed the error and updating votes was impossible
2) remove clearing cache on tx fail\cancellation, it's unnecessary and should be done only on success. Was added only for debugging

## Types of changes

### DIFF
--- a/apps/govern/components/Contracts/EditVotes/index.tsx
+++ b/apps/govern/components/Contracts/EditVotes/index.tsx
@@ -184,16 +184,6 @@ export const EditVotes = ({ allocations, setAllocations, setIsUpdating }: EditVo
         notification.error({
           message: error.message,
         });
-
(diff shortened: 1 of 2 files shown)


oaksprout (approved):

apps/govern/components/Contracts/EditVotes/index.tsx
```diff
@@ -184,16 +184,6 @@ export const EditVotes = ({ allocations, setAllocations, setIsUpdating }: EditVo
         notification.error({
           message: error.message,
         });
-
-        setIsUpdating(false);
```
truemiller, line 188: unsure if this links a loader further up the tree, is this no longer required?

truemiller (approved):

PR #71, opened 2024-07-30
(launch) feat: implement update

mohandast52:
## Types of changes

### DIFF
--- a/apps/launch/components/MyStakingContracts/Create/index.spec.tsx
+++ b/apps/launch/components/MyStakingContracts/Create/index.spec.tsx
@@ -168,7 +168,7 @@ describe('<CreateStakingContract />', () => {
   it('should display `Liveness period` field as required and able to fill the field', async () => {
     render(<CreateStakingContract />);
 
-    const livenessPeriodInput = screen.getByLabelText('Liveness period');
+    const livenessPeriodInput = screen.getByLabelText('Liveness period (sec)');
     expect(livenessPeriodInput).toBeRequired();
 
     expect(livenessPeriodInput).toHaveValue('86400'); // default value
@@ -178,7 +178,7 @@ describe('<CreateStakingContract />', () => {
   it('should display `Time for emissions` field as required, able to fill and see the error message when the value is out of range', async () => {
     render(<CreateStakingContract />);
 
-    const timeForEmissionsInput = screen.getByLabelText('Time for emissions');
+    const timeForEmissionsInput = screen.getByLabelText('Time for emissions (sec)');
     expect(timeForEmissionsInput).toBeRequired();
 
     expect(timeForEmissionsInput).toHaveValue('2592000'); // default value
--- a/apps/launch/components/MyStakingContracts/Details/index.spec.tsx
+++ b/apps/launch/components/MyStakingContracts/Details/index.spec.tsx
@@ -136,12 +136,12 @@ describe('<Details />', () => {
     },
     {
       testId: 'liveness-period',
-      title: 'Liveness period',
+      title: 'Liveness period (sec)',
       value: '86400 seconds',
     },
     {
       testId: 'time-for-emissions',
-      title: 'Time for emissions',
+      title: 'Time for emissions (sec)',
       value: '2592000',
     },
     {


apps/launch/components/MyStakingContracts/FieldConfig.tsx
```diff
@@ -0,0 +1,313 @@
...
+    maxNumInactivityPeriods: {
+      rules: getGenericFieldRules(FieldConfig.maxNumInactivityPeriods.name),
+    },
+    livenessPeriod: {
+      rules: [
+        ...getGenericFieldRules(FieldConfig.livenessPeriod.name),
+        {
+          type: 'number',
+          min: 86400,
+          max: 86400 * 30,
```
Tanya-atatakai, line 281: I don't see any min/max rules for this field anywhere, could you please tell me where you get them from?

apps/launch/components/MyStakingContracts/FieldConfig.tsx
```diff
@@ -0,0 +1,313 @@
...
+          message: `Liveness period must be between 24 hours and 30 days`,
+        },
+      ],
+    },
+    timeForEmissions: {
+      rules: [
+        {
+          required: true,
+          message: `Please enter Time for emissions`,
```
Tanya-atatakai, line 290: can use getGenericFieldRules here?

apps/launch/components/MyStakingContracts/Details/ContractConfiguration.tsx
```diff
@@ -0,0 +1,258 @@
...
+
+  const isLoading = useMemo(
+    () => isLivenessPeriodLoading || isMinimumStakingPeriodsLoading,
+    [isLivenessPeriodLoading, isMinimumStakingPeriodsLoading],
+  );
+  const data = useMemo(() => {
+    if (!livenessPeriod) return NA;
+    if (!minimumStakingPeriods) return NA;
+
+    const minStakingDuration = Number(minimumStakingPeriods) / Number(livenessPeriod);
```
Tanya-atatakai, line 96: I assume there's a confusion in naming here, should be `minimumStakingPeriods` = `minStakingDuration` / `livenessPeriod`

apps/launch/components/MyStakingContracts/Details/ContractConfiguration.tsx
```diff
@@ -0,0 +1,258 @@
...
+  const { data, isLoading } = useGetLivenessPeriod({ address });
+  return <ShowContent isLoading={isLoading} data={data ? `${data} seconds` : NA} />;
+};
+
+const TimeForEmissions: FC<{ address: Address }> = ({ address }) => {
+  const { data, isLoading } = useTimeForEmissions({ address });
+  return <ShowContent isLoading={isLoading} data={data} />;
+};
+
+const AgentInstances: FC<{ address: Address }> = ({ address }) => {
```
Tanya-atatakai, line 118: can you please rename this to `NumAgentInstances` as well as the `useAgentInstances` hook? currently it sounds like it returns a list of instances

apps/launch/components/MyStakingContracts/Details/ContractConfiguration.tsx
Tanya-atatakai: yes, the name of the component is `MinimumStakingPeriods` but here we're calculating and returning minStakingDuration which is another variable

kupermind:
@mohandast52 @Tanya-atatakai we are talking about time as hours, but setting the value in seconds. It didn't look nice when we tried to add (sec) to the Liveness period field, but something like this would be better. Otherwise we need to state that the period is at least 86400 seconds and at most 2... seconds.

kupermind (approved):

oaksprout:
@mohandast52 looks good but see my suggestion above for the comment.

Tanya-atatakai:
@mohandast52 I suggest we do the same for the `Time for emissions` field then, for consistency, because as Aleks mentioned it's also in seconds. So it'll be `Time for emissions (sec)`

Tanya-atatakai (approved):

oaksprout (approved):

PR #73, opened 2024-08-01
(govern) fix: update state after voting, fix rollover pool sorting, a…

Tanya-atatakai:
## Fixes

1) Fixed new votes didn't visually apply after voting (the state was cleared incorrectly)
2) Fixed rollover pool sorting (should always go at the end)
3) Add tooltip explaining "start of next week", fixed text

## Types of changes

### DIFF
--- a/apps/govern/components/Contracts/EditVotes/index.tsx
+++ b/apps/govern/components/Contracts/EditVotes/index.tsx
@@ -13,6 +13,7 @@ import { RETAINER_ADDRESS } from 'common-util/constants/addresses';
 import { INVALIDATE_AFTER_UPDATE_KEYS } from 'common-util/constants/scopeKeys';
 import { getBytes32FromAddress } from 'common-util/functions/addresses';
 import { voteForNomineeWeights } from 'common-util/functions/requests';
+import { NextWeekTooltip } from 'components/NextWeekTooltip';
 import { queryClient } from 'context/Web3ModalProvider';
 import { clearState } from 'store/govern';
 import { useAppDispatch, useAppSelector } from 'store/index';
@@ -218,7 +219,8 @@ export const EditVotes = ({ allocations, setAllocations, setIsUpdating }: EditVo
       />
 
       <Paragraph type="secondary" className="text-end">
-        New voting weight will take effect at the beginning of the next week.
+        Updated voting weights will take effect at the start of&nbsp;
+        <NextWeekTooltip hideDescription>next Unix time week.</NextWeekTooltip>
       </Paragraph>
       <Flex gap={12} justify="flex-end">
         <Button size="large" onClick={onCancel}>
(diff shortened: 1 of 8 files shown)


apps/govern/store/govern.ts
```diff
@@ -34,9 +34,7 @@ export const governSlice = createSlice({
     setLastUserVote: (state, action: PayloadAction<GovernState['lastUserVote']>) => {
       state.lastUserVote = action.payload;
     },
-    clearState: (state) => {
-      state = { ...initialState };
-    },
+    clearState: () => initialState,
```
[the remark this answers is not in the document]
Tanya-atatakai: the first one didn't work :D due to how the Immer library works under the hood of redux toolkit, I guess

kupermind (approved):

apps/govern/store/govern.ts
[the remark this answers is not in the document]
Tanya-atatakai: yeah, took me some time to realise

apps/govern/components/Contracts/EditVotes/index.tsx
```diff
@@ -218,7 +219,8 @@ export const EditVotes = ({ allocations, setAllocations, setIsUpdating }: EditVo
       />
 
       <Paragraph type="secondary" className="text-end">
-        New voting weight will take effect at the beginning of the next week.
+        Updated voting weights will take effect at the start of&nbsp;
```
oaksprout, line 222: `Updated voting weights will take effect at the start of next week (according to Unix time). <Tooltip>Learn more</Tooltip>`

apps/govern/components/NextWeekTooltip.tsx
```diff
@@ -10,14 +11,24 @@ const { Text } = Typography;
...
   return (
     <Tooltip
       color={COLOR.WHITE}
       overlayStyle={TOOLTIP_STYLE}
       title={
         <Space direction="vertical">
-          <Text>Updated voting weights will take effect at the start of next week Unix time.</Text>
+          {!hideDescription && (
+            <Text>
+              Updated voting weights will take effect at the start of next Unix time week.
```
oaksprout, line 29: `Updated voting weights will take effect at the start of next week (according to Unix time).`

oaksprout (changes requested):

apps/govern/components/Contracts/EditVotes/index.tsx
Tanya-atatakai: Talked to Roman and we agreed to remove the tooltip with the link from this state (where we are editing votes but haven't voted yet) because if no one voted during the week, the link would show an incorrect value, which might be confusing. So we just keep the text `"Updated voting weights will take effect at the start of next week (according to Unix time). "`

oaksprout (approved):

PR #76, opened 2024-08-12
(govern) display contract config fields on details page

Tanya-atatakai:
## Proposed changes

Added staking contract config fields on details page in Govern. It's copy-pasted from Launch with some updates in both:
- `useMaxNumServices` and `useRewardsPerSecond` and can be requested with help of `useStakingContractConstant`, moved it close to the other similar hooks
- in some fields values we repeated "seconds" or "OLAS" even thought we have it in labels already, removed that; also changed `(sec)` to `, seconds` so that they match the structure of other fields
- Updated / copy-pasted tests

I know it could be somehow a shared component in the libs folder, but the labels might change in the future, because they don't suit for the governor role

image

## Types of changes


oaksprout:
> Sorry, but I wouldn’t copy-paste

@mohandast52 makes sense to discuss that but I wouldn't want that to block the feature going live. If you agree to refactor that then let's do it in a follow up PR.

oaksprout (approved):

Tanya-atatakai:
@mohandast52 that's why I didn't create a separate lib - I was struggling with where to put it, both `common-contract-functions` and `ui-components` didn't look like a good fit. I don't think it makes sense to create an entire library just to store a few files that are similar across some apps. In the end, we risk having too many different lib folders, which could lead to confusion.

Maybe we can discuss this in our next engineering meeting? Generally, for this purpose, we could create something like the `shared` lib as in [nx-examples](https://github.com/nrwl/nx-examples/tree/master/libs/shared)

@truemiller would like to hear your thoughts too

DavidMinarsch (approved):

PR #85, opened 2024-08-27
(registry) fix: unitType to update hash for agents and components

mohandast52:
## Fixes

- The update hash for agents wasn’t working because the unitType was set to “0”. I’ve updated the code to address this and added a constant to prevent the use of hardcoded values in the future.​
- Removed all duplicate `UNIT_TYPES` across all the apps and added a single source of truth.

## Types of changes

### DIFF
--- a/apps/autonolas-registry/tests/components/ListAgents/details.test.jsx
+++ b/apps/autonolas-registry/tests/components/ListAgents/details.test.jsx
@@ -55,8 +55,8 @@ jest.mock('common-util/Details/utils', () => ({
   checkIfServiceRequiresWhitelisting: jest.fn(() => false),
 }));
 
-jest.mock('common-util/Details/DetailsSubInfo/hooks', () => ({
-  useTokenomicsUnitType: jest.fn(() => 1),
+jest.mock('common-util/Details/DetailsSubInfo/utils', () => ({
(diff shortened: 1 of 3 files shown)


apps/autonolas-registry/components/ListAgents/utils.jsx
```diff
@@ -72,7 +74,7 @@ export const updateAgentHashes = async (account, id, newHash) => {
   const contract = getMechMinterContract();
 
   // 0 to indicate `agents`
```
Tanya-atatakai, line 76: can we update the comment? or just remove it

truemiller (approved):

libs/util-constants/src/lib/units.ts
```diff
@@ -0,0 +1 @@
+export const UNIT_TYPES = { COMPONENT: '0', AGENT: '1' } as const;
```
Tanya-atatakai, line 1: we also have the `TOKENOMICS_UNIT_TYPES` in registry app, do you think we can have one source of truth? maybe we can keep rename this one to `TOKENOMICS_UNIT_TYPES`, remove old one from the app itself and update imports

Tanya-atatakai (approved):

PR #99, opened 2024-09-09
(bond) feat: support celo for bonding products

Tanya-atatakai:
### 1) adds path for celo-olas

image
image

### 2) supports celo chain on bonding products
Note: there are no celo products available atm, so I tested by changing config for current OLAS-WETH pair, got almost the same value - 146, when on prod it's 146.18

Screenshot 2024-09-09 at 19 44 10

## Types of changes

### DIFF
--- a/apps/bond/common-util/functions/web3.js
+++ b/apps/bond/common-util/functions/web3.js
@@ -1,4 +1,4 @@
-import { createPublicClient, getContract as getWagmiContract, http } from 'viem';
+import { createPublicClient, getContract as getViemContract, http } from 'viem';
 import Web3 from 'web3';
 
 import { CHAINS, RPC_URLS } from 'libs/util-constants/src';
@@ -61,7 +61,7 @@ export const getUniswapV2PairContractByChain = (address, chainId) => {
     transport: http(RPC_URLS[chainId] || chain.rpcUrls[0].http),
   });
 
-  const contract = getWagmiContract({
+  const contract = getViemContract({
     address: address,
     abi: UNISWAP_V2_PAIR_ABI,
     client: publicClient,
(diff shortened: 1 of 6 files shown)


apps/bond/common-util/functions/web3.js
```diff
@@ -1,17 +1,17 @@
+import { createPublicClient, getContract as getWagmiContract, http } from 'viem';
```
truemiller, line 1:
```suggestion
import { createPublicClient, getContract as getViemContract, http } from 'viem';
```
nitpicking here..  but wagmi just provides the hooks, viem provides the contract instance

apps/bond/components/BondingProducts/Bonding/BondingList/useBondingList.jsx
```diff
@@ -216,15 +252,10 @@ const useAddCurrentLpPriceToProducts = () => {
...
-            //   currentLpPricePromiseList.push(currentLpPricePromise);
-            // } else
-            if (dex === DEX.BALANCER) {
+            if (dex === DEX.UNISWAP) {
+              currentLpPrice = getCurrentPriceUniswap(productList[i].token);
+              otherRequests[i] = currentLpPrice;
+            } else if (dex === DEX.BALANCER) {
               currentLpPrice = getCurrentPriceBalancer(productList[i].token);
               otherRequests[i] = currentLpPrice;
             } else if (dex === DEX.SOLANA) {
```
truemiller, line 261: we should refactor the `DEX.###` keys at a later point..

using `ABI.UNISWAP_V2` for example, same for `.SOLANA`, would be better to identify the project and the type of contract we're interacting with

truemiller (approved):
looks good,
only nit pick comments, can dismiss for now

apps/bond/components/BondingProducts/Bonding/BondingList/utils.js
```diff
@@ -59,6 +60,10 @@ export const getSvmCalculatedPriceLp = (reserveOlas, totalSupply) => {
  * Function to get the link to the LP token
  */
 export const getLpTokenLink = ({ lpDex, lpChainId, lpPoolId, productName }) => {
+  if (lpChainId === celo.id) {
```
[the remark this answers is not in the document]
Tanya-atatakai: this one is correct, but should be similar in `getCurrentPriceLpLink`, thanks! updated

kupermind (approved):

PR #101, opened 2024-09-11
(chore) feat: Isolated deployments

mohandast52:
### This PR demonstrates isolated deployments, where only the affected apps should be re-deployed moving forward.

<hr />

- As shown in the screenshot below, the deployment was triggered only for the "govern" app due to the changes made only to the govern code, while the rest of the apps were skipped.

S2

- Each app will take approximately one minute to determine if it needs to be skipped during the Vercel deployment. The build log will show "Deployment has been canceled..." as determined by NX.

S3

S4

- For new apps, we need to update the "Ignored Build Step" in vercel and add a custom command using `npx nx-ignore <<app_name>>`. For example, for the "bond" app, we have `npx nx-ignore bond`.

Screenshot 2024-09-11 at 3 09 39 PM

- The screenshot below shows that **all** apps were triggered for deployment because changes were made in the `libs` folder. Changes in `libs/*` triggers deployment for all apps as it is a shared resource.

S6

### Time improvement 

- Previously, if a single app took an average of 3 minutes for deployment, deploying 5 apps would take approximately 15 minutes (5 apps × 3 minutes each).
- With the new approach, the deployment time should be reduced to approximately 7 minutes: (((number of apps - current app) x ~1 minute ) + deployment time of current app) i.e., ((5 - 1) * 1) + 3 = ~7 minutes.

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

### DIFF
--- a/libs/ui-theme/src/lib/ui-theme.spec.tsx
+++ b/libs/ui-theme/src/lib/ui-theme.spec.tsx
@@ -1,11 +1,11 @@
-import React from 'react';
 import { render } from '@testing-library/react';
+import React from 'react';
 
 import { AutonolasThemeProvider } from './ThemeConfig';
 
-describe('UiTheme', () => {
+describe('<AutonolasThemeProvider />', () => {
   it('should render successfully', () => {
-    const { baseElement } = render(<AutonolasThemeProvider  />);
+    const { baseElement } = render(<AutonolasThemeProvider />);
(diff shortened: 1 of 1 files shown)


truemiller (approved):

truemiller:
really nice 👏

Tanya-atatakai (approved):

PR #102, opened 2024-09-19
(operate) feat: update text "On Optimism"

mohandast52:
## Proposed changes

text

## Types of changes

### DIFF
--- a/apps/operate/components/Contracts/hooks.ts
+++ b/apps/operate/components/Contracts/hooks.ts
@@ -14,7 +14,7 @@ type StakingContractDetailsInfo = {
   availableOn?: StakingContract['availableOn'];
   minOperatingBalance?: number;
   minOperatingBalanceToken?: string;
-  isVariousChain?: boolean;
+  minOperatingBalanceHint?: string;
 };
 
 const STAKING_CONTRACT_DETAILS: Record<Address, StakingContractDetailsInfo> = {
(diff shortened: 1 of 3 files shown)


truemiller (approved):
seems fine, 
though I wonder if it's better to extend to with some popover that defines the chains?

i.e. [defillama's approach
](https://defillama.com/#:~:text=Pendle-,5%20chains,-Yield)image

oaksprout (changes requested):
The number amount also needs to be changed :) to ~0.17 ETH (Total Various Chains)

https://valory-workspace.slack.com/archives/C07852VAX51/p1726744358288159

apps/operate/components/Contracts/hooks.ts
```diff
@@ -55,9 +55,9 @@ const STAKING_CONTRACT_DETAILS: Record<Address, StakingContractDetailsInfo> = {
   },
   '0x00000000000000000000000088996bbde7f982d93214881756840ce2c77c4992': {
     availableOn: 'optimusQuickstart',
-    minOperatingBalance: 10,
+    minOperatingBalance: 0.17,
```
Tanya-atatakai, line 58: maybe could simplify it to just `minOperatingBalance: '~0.17 ETH'` and remove the `minOperatingBalanceToken ` since we don't need to convert the value (sad to delete such a nice code btw)
and maybe `isVariousChain` can be something like a `minOperatingBalanceHint: '(Total Various Chains)'` so the render logic is a bit easier to read... wdyt?

Tanya-atatakai (approved):

oaksprout (approved):

PR #105, opened 2024-09-25
(operate) feat: add blacklisted nominees

Tanya-atatakai:
## Proposed changes

Adds the list of addresses that shouldn't be shown on the operate app

## Types of changes

### DIFF
--- a/apps/operate/components/Contracts/hooks.ts
+++ b/apps/operate/components/Contracts/hooks.ts
@@ -6,7 +6,7 @@ import { useReadContracts } from 'wagmi';
 import { useNominees, useNomineesMetadata } from 'libs/common-contract-functions/src';
 import { RETAINER_ADDRESS } from 'libs/util-constants/src';
 import { STAKING_TOKEN } from 'libs/util-contracts/src';
-import { getAddressFromBytes32, getBytes32FromAddress } from 'libs/util-functions/src';
+import { areAddressesEqual, getAddressFromBytes32 } from 'libs/util-functions/src';
 
 const ONE_YEAR = 1 * 24 * 60 * 60 * 365;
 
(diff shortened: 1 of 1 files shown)


DavidMinarsch (approved):

truemiller (approved):

kupermind (approved):

DavidMinarsch:
Post-merge re-review — flagging follow-up items, not blockers.

**Correctness:** logic is right. `getAddressFromBytes32` strips to the last 40 hex chars and `areAddressesEqual` lowercases both sides, so checksum/case differences do not cause misses. No regression on the `RETAINER_ADDRESS` case.

**Follow-ups worth a small PR:**

- **No rationale for the two new addresses.** `0xaD9d…E029` and `0xE56d…7212` are inlined with no comment. A future maintainer cannot tell whether they are deprecated retainers, malicious nominees, or test contracts — which makes them dangerous to remove later. Add a one-line `// why hidden` per address.
- **Wrong layer.** `RETAINER_ADDRESS` lives in `libs/util-constants`; the two new siblings are inlined in an app hook. If govern/launch ever need the same hidden-nominee list, this duplicates. Recommend moving the list to `libs/util-constants/src`.
- **Naming.** `BLACKLISTED_ADDRESSES` reads as "malicious", but `RETAINER_ADDRESS` is a protocol address. Suggest `HIDDEN_NOMINEE_ADDRESSES` or `EXCLUDED_NOMINEES`. Also `blackListedNominee` → `blacklistedNominee` (one word).
- **Minor clarity.** Per nominee, the filter calls `getAddressFromBytes32(nominee.account)` then walks a 3-element array. Could precompute `BLACKLISTED_BYTES32 = BLACKLISTED_ADDRESSES.map(getBytes32FromAddress)` once and compare bytes32 directly. Negligible at N=3 but reads cleaner.
- **No tests.** Small filter, but worth a `.spec.ts` covering each address (mixed case + checksum), a non-blacklisted survivor, and an empty list.
- **App docs.** Per `CLAUDE.md`, app-level changes should update `apps/operate/CLAUDE.md`. A one-liner pointing at the constant would help.

**Security note for the PR description:** this is presentation-only filtering — anyone can patch the bundle. Fine if the intent is "don't show in UI", but if any of these must not be *operable on*, gating needs to be on-chain / server-side, not in the React filter.

PR #118, opened 2024-10-18
(govern): change formatting to local time

Tanya-atatakai:
## Proposed changes

Change time formatting from UTC to local (the updated function is also used on donate and veOLAS pages)

image

## Types of changes

### DIFF
--- a/apps/govern/components/VeOlas/VeOlasComponents.tsx
+++ b/apps/govern/components/VeOlas/VeOlasComponents.tsx
@@ -62,7 +62,7 @@ export const VotingPercentComponent = ({
     isLoading={isLoading}
     title="% of total voting power"
     value={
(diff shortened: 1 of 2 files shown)


apps/govern/components/VeOlas/VeOlasComponents.tsx
```diff
@@ -62,7 +62,7 @@ export const VotingPercentComponent = ({
     isLoading={isLoading}
     title="% of total voting power"
     value={
-      Number(votingPower) === 0 || Number(totalSupply) === 0
+      votingPower === undefined || Number(votingPower) === 0 || Number(totalSupply) === 0
```
Tanya-atatakai, line 65: just noticed we display null% for users with no veOLAS, quickly fixed

oaksprout (approved):

DavidMinarsch (approved):

PR #120, opened 2024-10-21
(launch, govern) feat: support mode network

Tanya-atatakai:
## Proposed changes

Support mode network on Govern and Launch. Tested on virtual test nets and forks, but with lots of troubles because of limitations.

1) Launch - contract creation

https://github.com/user-attachments/assets/e1cf26fd-e5a5-48b7-8e59-8630d4bab592

2) Launch - nominated contract
image

3) Govern - Voting

https://github.com/user-attachments/assets/d8fed8d3-b36a-46b6-b956-33e9a908e0e9

## Types of changes


apps/govern/components/Contracts/EditVotes/index.tsx
```diff
@@ -74,7 +74,7 @@ const getColumns = (
           value={allocations[index].weight}
           status={isError ? 'error' : undefined}
           onChange={(value) => {
-            if (value) {
+            if (typeof value === 'number') {
```
Tanya-atatakai, line 77: this fixes the ability to reset existing votes to zero @kupermind you complained about that it returns the value to the previous one

apps/launch/common-util/constants/stakingContract.ts
```diff
@@ -51,6 +52,7 @@ export const IMPLEMENTATION_ADDRESSES: Addresses = {
   [base.id]: '0xEB5638eefE289691EcE01943f768EDBF96258a80',
   [arbitrum.id]: '0x04b0007b2aFb398015B76e5f22993a1fddF83644',
   [celo.id]: '0xe1E1B286EbE95b39F785d8069f2248ae9C41b7a9',
+  [mode.id]: '0xE49CB081e8d96920C38aA7AB90cb0294ab4Bc8EA',
```
Tanya-atatakai, line 55: @kupermind would be great of you could check all added addresses on this page

libs/util-contracts/src/lib/abiAndAddresses/stakingFactory.js
```diff
@@ -10,6 +10,7 @@ export const STAKING_FACTORY = {
     [base.id]: '0x1cEe30D08943EB58EFF84DD1AB44a6ee6FEff63a',
     [arbitrum.id]: '0xEB5638eefE289691EcE01943f768EDBF96258a80',
     [celo.id]: '0x1c2cD884127b080F940b7546c1e9aaf525b1FA55',
+    [mode.id]: '0x75D529FAe220bC8db714F0202193726b46881B76',
```
Tanya-atatakai, line 13: @kupermind and also the factory address

apps/launch/common-util/constants/stakingContract.ts
kupermind: Just FYI this is confusing for us :). We call this `StakingTokenAddress` :).

apps/launch/common-util/constants/stakingContract.ts
```diff
@@ -81,6 +85,7 @@ export const STAKING_TOKEN_ADDRESSES: Addresses = {
   [base.id]: '0x54330d28ca3357F294334BDC454a032e7f353416',
   [arbitrum.id]: '0x3d77596beb0f130a4415df3D2D8232B3d3D31e44',
   [celo.id]: '0xaCFfAe8e57Ec6E394Eb1b41939A8CF7892DbDc51',
+  [mode.id]: '0xcfD1D50ce23C46D3Cf6407487B2F8934e96DC8f9',
```
kupermind, line 88: And we all this `olasAddress` :P

kupermind (approved):

DavidMinarsch (approved):

apps/launch/common-util/constants/stakingContract.ts
Tanya-atatakai: yeah, I was also confused when updating these constants, renamed :D

truemiller:
bit overkill to define new chains with the same chainIds as the origin fork :/
could overwrite the RPCs if on development and that would probably be enough
no biggy though

truemiller (approved):

Tanya-atatakai:
> bit overkill to define new chains with the same chainIds as the origin fork :/
could overwrite the RPCs if on development and that would probably be enough
no biggy though

@truemiller good point actually, will consider changing that. but there's also something I'd like to discuss on our next eng call re this.

PR #135, opened 2024-11-14
chore: added title to Veolas for duplicate title error in SEMrush

atepem:
## Proposed changes

Just added a title to VeOlas to resolve SEMrush issues, duplicate title tags:
image

### DIFF
--- a/apps/govern/components/VeOlas/index.tsx
+++ b/apps/govern/components/VeOlas/index.tsx
@@ -1,14 +1,12 @@
-import { Alert, Button, Card, Space, Typography } from "antd";
-import { useState } from "react";
-import styled from "styled-components";
+import { Alert, Button, Card, Space, Typography } from 'antd';
+import { useState } from 'react';
+import styled from 'styled-components';
 
-import { useFetchBalances } from "hooks/index";
+import { useFetchBalances } from 'hooks/index';
 
-import { CreateLockModal } from "./CreateLockModal";
-import { IncreaseLockModal } from "./IncreaseLockModal";
-import { VeOlasManage } from "./VeOlasManage";
-import { GetServerSideProps } from "next";
-import Head from "next/head";
+import { CreateLockModal } from './CreateLockModal';
+import { IncreaseLockModal } from './IncreaseLockModal';
+import { VeOlasManage } from './VeOlasManage';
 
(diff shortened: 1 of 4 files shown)


apps/govern/components/VeOlas/index.tsx
```diff
@@ -23,53 +25,64 @@ const Title = styled.h1`
...
-                setIsIncreaseModalVisible(true);
-              } else {
-                setIsCreateLockModalVisible(true);
-              }
-            }}
-          >
-            Lock OLAS for veOLAS
-          </Button>
+    <>
+      <Head>
```
Tanya-atatakai, line 37: could you please add the whole meta block to govern app similar to one that we have on olas website or contribute or in bond app here apps/bond/pages/_app.jsx? currently govern app is not searchable in google :( I probably forgot to add that when created the app

Tanya-atatakai (changes requested):
1) need to fix linter, github actions are failed
2) don't understand what it means "duplicated titles" in SEMrush, currently there's only one on veOLAS page 🤔 
3) we didn't fix it for the contracts page, is that expected?

apps/govern/components/VeOlas/index.tsx
```diff
@@ -23,53 +25,64 @@ const Title = styled.h1`
...
+        <title>VeOlas</title>
```
Tanya-atatakai, line 38: should be `veOLAS`

Tanya-atatakai (changes requested):
1) need to fix linter, github actions are failed
2)

apps/govern/pages/veolas.tsx
```diff
@@ -1,3 +1,15 @@
-import { VeOlasPage } from '../components/VeOlas';
+import Meta from 'components/Meta';
+import { VeOlas } from '../components/VeOlas';
+
+const VeOlasPage = () => (
+  <>
+    <Meta
+      pageTitle="veOlas"
```
Tanya-atatakai, line 7:
```suggestion
      pageTitle="veOLAS"
```

Tanya-atatakai (approved):
a small comment, approving in advance, otherwise looks very good! thanks

oaksprout (approved):

PR #160, opened 2025-04-01
fix: safe interactions

Tanya-atatakai:
## Fixes

Possibly fixes issues with safe interactions

## Types of changes

### DIFF
--- a/apps/govern/common-util/config/wagmi.ts
+++ b/apps/govern/common-util/config/wagmi.ts
@@ -1,4 +1,4 @@
-import { createConfig, http } from 'wagmi';
+import { cookieStorage, createStorage } from 'wagmi';
 import {
   Chain,
   arbitrum,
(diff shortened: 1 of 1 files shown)


apps/govern/common-util/config/wagmi.ts
```diff
@@ -27,30 +27,27 @@ export const SUPPORTED_CHAINS: [Chain, ...Chain[]] = [
...
-    coinbaseWallet({
-      appName: walletConnectMetadata.name,
-    }),
-  ],
-  transports: SUPPORTED_CHAINS.reduce(
-    // TODO: merge two sources of RPCs into one?
-    (acc, chain) => Object.assign(acc, { [chain.id]: http(RPC_URLS[chain.id]) }),
-    {},
-  ),
+export const wagmiConfig = defaultWagmiConfig({
```
Tanya-atatakai, line 48: had to use `defaultWagmiConfig` because it seems we should define safe connector somehow differently. works in registry, so just copied from there

PR #174, opened 2025-06-25
feat: build application workflow

rajat2502:
## Proposed changes

Build workflow for the `build` application

## Types of changes


.github/workflows/build.yml
```diff
@@ -0,0 +1,41 @@
+# TODO: figure out how to trigger workflow from vercel for nx
```
Tanya-atatakai, line 1: I think these comments can be removed, it was related to ignoring other apps and building only changed onex

PR #205, opened 2025-08-25
(marketplace) fix: marketplace improvements & fixes

rajat2502:
## Proposed changes

Adds better types, constant changes, and fixes from previous PR reviews

## Types of changes


apps/marketplace/common-util/Details/ActivityDetails.tsx
```diff
@@ -1,12 +1,12 @@
 import type { Activity } from 'common-util/graphql/service-activity';
-import { Flex, Modal, Typography } from 'antd';
+import { Col, Flex, Modal, Row, Typography } from 'antd';
 import { AddressLink, NA } from '@autonolas/frontend-library';
```
Tanya-atatakai, line 3: it would be nice to use the existing AddressLink with some extensions (the ability to copy), the on in this library doesn't have correct scan link for mode. can you possibly fix this as part of this cleanup?

apps/marketplace/common-util/Details/index.tsx
```diff
@@ -56,8 +47,8 @@ const getColumns = ({
             onClick={() => openActivityModal(record)}
             style={{ display: 'flex', gap: 8 }}
           >
-            {<AddressLink suffixCount={8} textMinWidth={160} text={text} cannotClick />}
-            <CopyBtn text={text} />
+            <AddressLink suffixCount={8} textMinWidth={160} text={text} cannotClick />
+            <Copy text={text} />
```
Tanya-atatakai, line 51: wonder, why this wasn't initially done via "canCopy" prop in AddressLink

apps/marketplace/components/ListServices/utils.tsx
```diff
@@ -133,14 +134,15 @@ export const getServices = async (
 
   let servicesWithMetadata = await extractConfigDetailsForServices(results);
   if (chainId === 100 || chainId === 8453) {
```
Tanya-atatakai, line 136: maybe could introduce a small helper that based on chainId returns true or false if marketplace is enabled? also can be used here `apps/marketplace/common-util/Details/index.tsx` and `apps/marketplace/pages/api/services.ts`

apps/marketplace/common-util/Details/index.tsx
rajat2502: the styles weren't matching with figma, didn't wanted to override

Tanya-atatakai (approved):

PR #353, opened 2026-04-20
chore: (BDMARK-2086) package.jsons + SUPPLY-CHAIN-SECURITY.md

atepem:
## Proposed changes

* Updated package.jsons + yarn.lock
* Added SUPPLY-CHAIN-SECURITY.md

## Summary
Hardens the monorepo against npm supply-chain attacks. Three pieces:

1. **Pin everything.** All direct deps (root + libs) pinned to exact versions. 28 Yarn `resolutions` pin transitive deps with high/critical CVEs.
2. **CI gates.** Three blocking jobs in [supply-chain.yml](.github/workflows/supply-chain.yml): `yarn audit:prod` (production-tree audit with allowlist), `lockfile-lint` (registry/HTTPS/integrity), `audit:install-hooks` (diff postinstall surface against allowlist). All actions SHA-pinned across 6 app workflows. `SNYK_TOKEN` scoped per-step in [snyk-security.yml](.github/workflows/snyk-security.yml) — closes a postinstall-exfil window.
3. **Doc.** [SUPPLY-CHAIN-SECURITY.md](SUPPLY-CHAIN-SECURITY.md) covers threat model, policies, and response playbook.

## Audit delta
**137 unique advisories → 0 critical, 3 high allowlisted** (`bigint-buffer`, `next` ×2). All allowlist entries carry reasons + 3-month review dates in [.supply-chain/audit-allowlist.json](.supply-chain/audit-allowlist.json).

## Removed
- `ipfs@0.63.5` — direct dep, not imported anywhere. Pulled in `electron@1.8.8` + `wrtc` + a `headless`-from-tarball with no integrity hash. Removal cleared the `lockfile-lint` failure and two postinstall binary downloads.

## Breaking-change risks
Major-version `resolutions` that force consumers onto newer APIs:
- `tar: 4 → 7` — consumer: `swarm-js@^0.1.40` via `web3 → web3-bzz`
- `undici: 5 → 7` — consumers: `@vercel/blob@0.23.4`, `dns-over-http-resolver@^2.1.0`
- `serialize-javascript: 6 → 7` — consumers: `copy-webpack-plugin`, `terser-webpack-plugin`

Verified by `yarn nx run-many --target=build --all` — all 12 projects build clean. `minimatch@9` / `glob@10` were rolled back for the same class of risk — see doc.

Direct-dep bumps bundled in: `axios 1.2.1 → 1.13.5`, `node-fetch 2.0.0 → 2.7.0`, `lodash 4.17.21 → 4.18.1` (§5 cooldown checked: published 2026-04-01, 26 days old), `isomorphic-dompurify 2.9 → 2.36.0`, `antd 5.9 → 5.16.4`, `@particle-network/auth 1.2.2 → 1.3.1`, `@reduxjs/toolkit 2.0.1 → 2.2.3`.

## Test plan
- [x] `yarn install --frozen-lockfile`
- [x] `yarn audit:prod` — 3 allowlisted, no unlisted high/critical
- [x] `yarn lint:lockfile`
- [x] `yarn audit:install-hooks`
- [x] `yarn nx run-many --target=build --all` — all 12 projects pass
- [ ] Manual smoke of apps using `@vercel/blob` (OG image generation) and `web3-bzz`

## Out of scope (tracked in doc TODOs)
- Next.js 14 → 15 migration — clears 2 of the 3 allowlisted highs
- `gitleaks.yml` SHA pins
- `operate.yml` lacking `yarn test` (pre-existing)

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)


atepem:
## Summary

Harden the monorepo against npm supply-chain attacks. Pins direct and a
few transitive deps, removes an unused-but-postinstall-heavy `ipfs` chain,
clears several direct-dep CVEs, and adds three blocking CI gates
(`yarn audit`, `lockfile-lint`, install-hook allowlist diff) wired through
a new `.github/workflows/supply-chain.yml`. Full policy and response
playbook documented in `SUPPLY-CHAIN-SECURITY.md`.

## Changes

### Documentation
- **New** `SUPPLY-CHAIN-SECURITY.md` — threat model, 9 policies, response
  playbook, active TODO list. Complements `SECURITY.md`.

### CI gates (all blocking, SHA-pinned)
- **New** `.github/workflows/supply-chain.yml` — three jobs:
  - `yarn audit --groups dependencies` with Yarn 1.x exit-code bitmask
    gating (fail on `high | critical`, i.e. `exit_code & 24`).
  - `lockfile-lint` — validates every `resolved` URL is on
    `registry.yarnpkg.com` / `registry.npmjs.org`, uses HTTPS, has an
    integrity hash.
  - Install-hook audit — diffs non-trivial `pre/install/post` scripts
    against `.supply-chain/install-hooks.allowlist`; fails on drift.
- **New** `scripts/audit-install-hooks.mjs` + `.supply-chain/install-hooks.allowlist`
  (14 baseline entries, depth-bounded recursion).
- **Modified** app workflows (`build`, `contribute`, `govern`, `launch`,
  `marketplace`, `operate`): `yarn` → `yarn install --frozen-lockfile`;
  SHA-pinned `actions/checkout@v4.1.7`, `actions/setup-node@v4.0.3`,
  `actions/cache@v4.2.0`. Marketplace also bumped `setup-node@v3` → `v4`.
- **Modified** `snyk-security.yml`: `SNYK_TOKEN` moved from job-level
  `env:` to per-step `env:` on Snyk steps only — closes the
  postinstall-exfil window during `yarn install`.

### Dependency hygiene
- `packageManager: "yarn@1.22.22"` pinned in root `package.json`.
- `package-lock.json` / `pnpm-lock.yaml` added to `.gitignore`.
- Exact-pin all direct deps in root `package.json` and `libs/*/package.json`.
- **Removed** `ipfs@0.63.5` (never imported — only `ipfs-http-client` is
  used). Kills `electron@1.8.8` + `wrtc@0.4.7` + `headless`-from-tarball
  chain; unblocks `lockfile-lint`. `yarn.lock` shrinks by ~3.4k lines.
- `axios` `1.2.1` → `1.7.7` (CVE-2023-45857, CVE-2024-39338).
- `node-fetch` `2.0.0` → `2.7.0` (CVE-2020-15168, CVE-2022-0235).
- Yarn `resolutions`:
  - `@balancer-labs/sdk/axios` → `1.7.7` (transitive axios).
  - `form-data` → `2.5.5` globally (CVE-2024-42459; 2.x/3.x/4.x consumer
    ranges collapse to one entry).
- New `package.json` scripts: `audit`, `audit:install-hooks`,
  `audit:install-hooks:update`, `lint:lockfile`.
- New devDep: `lockfile-lint@5.0.0`.

## Test plan

Local gates all green:
- [x] `yarn install --frozen-lockfile` — clean.
- [x] `yarn lint:lockfile` — ✔ No issues detected.
- [x] `yarn audit:install-hooks` — OK (14 allowlisted).
- [x] `yarn why form-data` — all 11 consumer paths resolve to `2.5.5`.
- [x] `yarn why axios` — `@balancer-labs/sdk#axios` resolves to `1.7.7`.

CI checks to watch on first run:
- [x] Supply Chain / `yarn audit` job passes (no high/critical in prod tree).
- [x] Supply Chain / `lockfile-lint` job passes.
- [x] Supply Chain / install-hook audit job passes.
- [ ] All six app workflows pass with `--frozen-lockfile`.
- [ ] Snyk workflow still completes (SARIF upload unchanged).

## Follow-ups (tracked in `SUPPLY-CHAIN-SECURITY.md` TODO)

- SHA-pin `.github/workflows/gitleaks.yml` (still on `@v3`).
- Reserve `@autonolas-frontend-mono` on npm or rename to an owned scope.
- **Audit Vercel env-var scoping per app** — every secret (RPC URLs,
  API keys, Blob tokens) must be marked runtime-only, not build-time.
  Biggest real-world gap between policy and practice.
- Verify each app's Vercel install command is `yarn install --frozen-lockfile`.

Tanya-atatakai:
# PR Review

## Overview

A supply-chain hardening pass. It pins every direct/devDependency to exact versions, adds 28 transitive `resolutions` to clear ~134 high/critical CVEs, removes the unused `ipfs` direct dep (kills the `electron@1.8.8` + `wrtc` + `headless`-from-tarball chain), and introduces a CI gate (`yarn audit:prod`, `lockfile-lint`, `install-hook audit`). It also SHA-pins all reusable GitHub Actions in app workflows, scopes `SNYK_TOKEN` per-step (not job-wide), and adds a thorough `SUPPLY-CHAIN-SECURITY.md` documenting the threat model and policies. Net diff is `+1599 / −4522` (most of the deletions come from removing `ipfs`'s transitive tree from `yarn.lock`).

## What's strong

- **The doc earns its keep.** `SUPPLY-CHAIN-SECURITY.md` records *why* every choice was made (e.g., why `audit:prod` not `audit` — Yarn 1.x's stock `yarn audit` shadows same-named scripts; why `minimatch@5.1.8` not `9.0.7` — `nx` depends on the callable default export removed in 9.x; why no `glob@10.5.0` resolution — tree has no 10.x consumers). Future readers won't have to reverse-engineer the resolution list.
- **`scripts/audit.mjs` is correctly belt-and-braces.** Yarn 1.x `yarn audit` exits with a *severity bitmask*, not a threshold, so wrapping it and parsing JSON is the right call. The blocking/suppressed/expired/stale partitioning is clean and the `::warning::` on expired-but-still-allowlisted entries (vs. failing) is the right ergonomic choice — the team gets surfaced reminders without hard merge blocks.
- **`install-hooks.allowlist` as a diff-driven review surface.** Turning "a new package with a postinstall just entered my dep graph" into an explicit allowlist diff is exactly the right mechanism. The `TRIVIAL` filter (`echo`, `true`, `:`, `exit 0`) is restrained — it intentionally keeps `try/catch` shims like `es5-ext`'s `node -e "try{require(...)}catch{}"` in scope, which matches the threat model.
- **`SNYK_TOKEN` step-scoping** in `.github/workflows/snyk-security.yml` closes a real exfil window. Comment in the workflow explains the threat directly.
- **`.gitignore` blocks `package-lock.json` / `pnpm-lock.yaml`.** Stops the "stray `npm install` lands a second lockfile" footgun.

## Issues / risks

### 1. Major-version `resolutions` for `tar`, `undici`, `serialize-javascript` are aggressive

In `package.json` the resolutions force:
- `tar: 7.5.11` while consumers request `tar@^4.0.2` (4 → 7 is a major-major jump; v7 dropped CommonJS sync APIs in favor of the new promise interface)
- `undici: 7.24.0` while consumers request `^5.12.0` / `^5.28.4` (`fetch` API surface changed; `Dispatcher` API moved)
- `serialize-javascript: 7.0.3` while consumers request `^6.0.0` / `^6.0.1`

The doc acknowledges this kind of pitfall for `minimatch@9` and `glob@10` and rolled them back, but doesn't call out that the same risk applies to `tar`/`undici`/`serialize-javascript`. **Suggestion:** verify build/runtime smoke-test passes for every app (`yarn nx run-many --target=build --all`) before merge, and add a one-liner in the doc next to the `minimatch`/`glob` notes for these three explaining why the major bump is safe in this tree.

### 2. `lodash@4.18.1` is real but unusual

The diff bumps `lodash` from `^4.17.21` → `4.18.1`. `4.17.21` was the long-standing canonical and `4.18.x` was published recently. I confirmed `4.18.1` exists on npm — but it's the kind of version where the §5 cooldown rule (≥7 days old) and §9 dependency review (publish history, maintainer, transferred ownership) are most important. The PR description doesn't note whether the cooldown was checked. **Suggestion:** in the PR description, paste `npm view lodash@4.18.1 time` output (or just confirm the cooldown was checked), per §5 of the very doc this PR adds.

### 3. `audit.mjs` keys advisories on numeric `id`, not GHSA

`scripts/audit.mjs` and `.supply-chain/audit-allowlist.json` both key on the npm-advisory numeric `id` (`1103747`, `1112653`, `1116376`). The numeric IDs are stable in practice but the GHSA IDs (`GHSA-3gc7-fjrx-p6mg`, etc.) are the canonical, vendor-agnostic identifiers and would survive a registry-side renumbering or migration. The allowlist already carries `ghsa` as a "for humans" field — consider matching on either field (prefer GHSA, fall back to numeric `id`) so the allowlist isn't tied to npm's internal IDs. Not a blocker.

### 4. PR description is too thin for a 6121-line diff

The description is just "Updated package.jsons + yarn.lock / Added SUPPLY-CHAIN-SECURITY.md". For a diff that pins ~80 dependencies, adds 28 `resolutions`, introduces 3 CI jobs, and rewrites SHA pins across 6 workflows, the reviewer needs the *summary* the doc already contains: the "starting point was 137 advisories → 0 critical, 3 high allowlisted" line, the `ipfs` removal rationale, and the breaking-change risks. **Suggestion:** copy the relevant TODO checkboxes from `SUPPLY-CHAIN-SECURITY.md` into the PR body.

### 5. Minor: a bunch of *non-pin* version bumps are bundled in

The PR title says "package.jsons + yarn.lock" but it's also a real upgrade: `axios 1.2.1 → 1.13.5`, `node-fetch 2.0.0 → 2.7.0`, `isomorphic-dompurify ^2.9.0 → 2.36.0`, `antd ^5.9.0 → 5.16.4`, `@particle-network/auth ^1.2.2 → 1.3.1`, `@reduxjs/toolkit ^2.0.1 → 2.2.3`, etc. Most are CVE-driven and the doc lists the CVE-driven ones explicitly. The non-CVE bumps (e.g., `antd`, `@particle-network/auth`) appear to be incidental — the author probably regenerated `yarn.lock` and pinned the resolved version. That's defensible but worth calling out: the PR is doing two things (pin + bump) and a future bisect across this commit will be wider than the title suggests.

### 6. Minor: `operate.yml` doesn't run tests

Not introduced by this PR (pre-existing), but visible in the diff — the workflow has only ESLint, no `yarn test`. Out of scope here, but flag-worthy as a separate ticket.

## Security considerations

- **Net positive.** The blocking gates (`yarn audit:prod`, `lockfile-lint`, `install-hook audit`) materially raise the bar for a future supply-chain attack landing silently. The job-wide → step-scoped `SNYK_TOKEN` change in particular closes a real (if narrow) postinstall-exfil window.
- **One residual gap is acknowledged in the doc** (`gitleaks.yml` floating `@v3` tags). Tracked as an unchecked TODO. Fine to defer to a follow-up PR.
- **No new secrets, tokens, or credentials introduced.** Scripts read only `node_modules/` and `.supply-chain/` files — paths are anchored to `resolve('.')`, no `argv`-derived path sinks.

## Recommendation

**Approve with the following before merge:**
1. Confirm `yarn nx run-many --target=build --all` and a smoke run of each app passes — the major-version resolution bumps for `tar`/`undici`/`serialize-javascript` are the highest-risk part of this diff.
2. Beef up the PR description with the summary already in the doc (advisory count delta, breaking-change list, removed deps).
3. Confirm §5 cooldown was checked for `lodash@4.18.1` (and ideally pasted into the PR description).

The doc and CI gates are excellent. The risk is concentrated in transitive resolution behavior, not the policy framework itself.

rajat2502:
Reviewed. The policy + mechanism work here is great — `audit.mjs`, `audit-install-hooks.mjs`, and `SUPPLY-CHAIN-SECURITY.md` are solid.

Two items already raised by prior reviewers that I'd like to see addressed before merge:

- **@Tanya-atatakai's point on major-version `resolutions`** — `tar` (4→7), `undici` (5→7), `serialize-javascript` (6→7) carry the same API-breakage risk class as the `minimatch@9` / `glob@10` rollbacks the doc explicitly calls out, but aren't flagged. Suggest `yarn nx run-many --target=build --all` + a manual boot of each app to confirm runtime behaviour, and a one-liner in the doc next to the minimatch/glob notes explaining why the major bump is safe for these three.
- **Copilot's `scripts/audit.mjs:109` stale-detection fix** — one-line bug: if a suppressed advisory's severity drops below high/critical, the allowlist entry never gets flagged as stale. The suggested patch (compare against the `suppressed` set instead of all advisories) is correct.

Everything else (GHSA vs numeric `id`, allowlist required-field validation, PR description thinness, `operate.yml` lacking tests) is nit / follow-up material — fine to defer.

mohandast52:
# Review – PR #353 


## TL;DR

The doc and CI gates are excellent. Net-positive on supply-chain posture (137 advisories → 0 critical, 3 allowlisted highs; install-hook surface now diff-gated; SHA-pinned actions; per-step `SNYK_TOKEN`).

The risk is concentrated in a handful of major-version `resolutions` that force transitive consumers onto APIs they were not written against. **Approve once those are smoke-tested or rolled back, and once the small `audit.mjs` tightening is in.**

---

## What's strong

- **`SUPPLY-CHAIN-SECURITY.md` carries its weight.** Reasoning is captured in-place: why `audit:prod` not `audit` (Yarn 1.x's built-in shadows same-named scripts), why `minimatch@5.1.8` not `9.0.7` (nx needs the callable default export), why no `glob@10.5.0` resolution (no 10.x consumers in the tree). Future readers won't have to reverse-engineer the resolutions block.
- **`scripts/audit.mjs` correctly handles Yarn 1.x's exit-code bitmask** by parsing JSON and applying its own gate. `blocking / suppressed / expired / stale` partitioning is clean. Using `::warning::` (not failure) for expired-but-still-allowlisted entries is the right ergonomic — surfaces reminders without blocking merges.
- **Install-hook allowlist as a diff-driven review surface** is the right mechanism. The `TRIVIAL` filter (`echo`, `true`, `:`, `exit 0`) intentionally keeps `try/catch` shims like `es5-ext`'s in scope, which matches the threat model.
- **`SNYK_TOKEN` step-scoping** in `snyk-security.yml` closes a real exfil window during `yarn install`, and the inline comment explains why.
- **`.gitignore` blocks `package-lock.json` / `pnpm-lock.yaml`** — kills the "stray `npm install` lands a second lockfile" footgun.
- **`packageManager: yarn@1.22.22`** + corepack activation in CI means we're not at the mercy of whatever Yarn the runner ships with.

---

## Must address before merge

### 1. Major-version `resolutions` for `tar`, `undici`, `serialize-javascript` — verify, don't trust

These three resolutions force consumers onto APIs that were never tested against:

| Resolution | Forces | Real consumers requesting older majors |
|---|---|---|
| `tar: 7.5.11` | 4.x → 7.x | `swarm-js@^0.1.40` requests `tar@^4.0.2` (reached via `web3 → web3-bzz → swarm-js`) |
| `undici: 7.24.0` | 5.x → 7.x | `@vercel/blob@0.23.4` requests `undici@^5.28.4`; `dns-over-http-resolver@^2.1.0` requests `undici@^5.12.0` |
| `serialize-javascript: 7.0.3` | 6.x → 7.x | `copy-webpack-plugin@^10.2.4`, `terser-webpack-plugin@^5.3.10` request `^6.0.0` / `^6.0.1` |

`tar 4 → 7` dropped the sync CommonJS APIs in favour of promise APIs. `undici 5 → 7` reshuffled the `Dispatcher` and fetch surface. `serialize-javascript 6 → 7` is the lowest-risk of the three.

The `@vercel/blob` one is the highest concern in production — OG image / blob storage paths in any app that uses it. The `swarm-js → tar` path is exercised via `web3-bzz` in the `web3` chain.

The doc already records the same class of pitfall for `minimatch@9` and `glob@10` and rolled them back. Apply the same discipline here:

- [ ] Run `yarn nx run-many --target=build --all` (full monorepo build, all 9 apps).
- [ ] Smoke each app — at minimum the apps using `@vercel/blob` (OG image generation) and any code path touching `web3-bzz`.
- [ ] If anything regresses, narrow the resolution (e.g. `tar` to the latest `4.x` patch, `undici` to the latest patched `5.x`) instead of holding the major.
- [ ] Whatever lands, add a one-line "why the major bump is safe in this tree" note to `SUPPLY-CHAIN-SECURITY.md` next to the existing `minimatch` / `glob` notes.

### 2. `lodash@4.18.1` — confirm §5 cooldown was checked

The bump is `^4.17.21` → `4.18.1`. `4.17.21` was the long-standing canonical and `4.18.x` was published recently. `4.18.1` exists on npm and is the kind of version where this PR's own §5 (≥7-day cooldown) and §9 (publish history, maintainer, transferred ownership) matter most.

- [ ] Paste `npm view lodash@4.18.1 time` (or just confirm the cooldown was checked) into the PR description, per §5 of the doc this PR adds.

### 3. `scripts/audit.mjs` — small correctness fixes

Both Copilot's review comments are valid:

- **Allowlist entries with missing `reason` / `review` silently suppress and never expire.**  Header doc says every entry needs both, but `loadAllowlist` keys on `typeof entry.id === 'number'` only. A future PR that drops `reason` (typo, copy-paste) suppresses the advisory with no warning. Validate required fields and fail fast (`process.exit(2)`) on a malformed entry.

- **"Stale" check misclassifies severity drops.**  An allowlisted entry is marked stale only when its advisory ID is completely absent from the audit output. If the advisory is still present but its severity drops below high/critical, the entry is no longer suppressing anything and should be flagged as stale. Compare against the `suppressed` set, not the full advisory list:

  ```js
  // Entries allowlisted but no longer suppressing a high/critical advisory — drift.
  const suppressedIds = new Set(suppressed.map(({ advisory }) => advisory.id));
  for (const entry of allowlist.entries || []) {
    if (!suppressedIds.has(entry.id)) {
      stale.push(entry);
    }
  }
  ```

### 4. PR description is too thin for a 6121-line diff

For a diff that pins ~80 deps, adds 28 transitive `resolutions`, introduces 3 CI jobs, and SHA-pins actions across 6 workflows, a reviewer needs the summary already in the doc:

- [ ] Copy the "137 advisories → 0 critical / 3 allowlisted highs" delta from the SUPPLY-CHAIN-SECURITY.md TODO list into the PR body.
- [ ] Call out the breaking-change risks (the three majors above + any incidental upgrade like `antd 5.9 → 5.16.4`, `axios 1.2.1 → 1.13.5`, `@particle-network/auth 1.2.2 → 1.3.1`, `@reduxjs/toolkit 2.0.1 → 2.2.3`, `isomorphic-dompurify 2.9 → 2.36.0`).
- [ ] List the removed deps (`ipfs@0.63.5`) and why.

A future bisect across this commit will be wider than the title suggests — the PR is doing two things (pin + bump) and the description should make both visible.

---

## Nice-to-have (don't block)

### 5. SAST findings on `audit-install-hooks.mjs` should be marked resolved

The three GitHub Advanced Security path-traversal findings (lines 94, 111, 142) were posted at `2026-04-24T16:10Z` against the `argv[2]`-as-root variant. Commit `fac621068` (`chore: unsanitised inputs`, `2026-04-24T19:03Z`) replaced `process.argv[2]` with `resolve('.')` and added an inline comment explaining why argv is not accepted. The findings are obsolete on the latest commit — confirm Snyk's next scan dismisses them, then resolve in the GHAS UI.

### 6. Install-hook allowlist matches by name only

The allowlist tracks **names**, with the hook command captured only as a comment. A malicious update to an allowlisted package (e.g. a hijacked `@swc/core`) that swaps its `postinstall` from `node postinstall.js` to `curl … | sh` would not trigger the diff. The threat model (§3 in the doc — "malicious publish to a legitimate name") explicitly covers this case, but the install-hook gate doesn't.

Cheap upgrade for a follow-up PR: hash or fully-string-match the hook command set per allowlisted name, and fail when a name's command set changes. Stays a 1-file script, doesn't expand the workflow.

### 7. Allowlist matching by numeric `id`

Already raised by Tanya. The npm advisory numeric IDs are stable in practice but the `ghsa` field is the canonical, vendor-agnostic identifier. The allowlist already carries `ghsa` — match on either (prefer GHSA, fall back to numeric `id`). Not a blocker. Defer to a follow-up unless you're already in `audit.mjs` for §3.

### 8. `audit.mjs` — minor

`const { stdout, stderr, code } = await runYarnAudit();` destructures `code` but never uses it. Drop it.

### 9. `operate.yml` doesn't run `yarn test`

Pre-existing — visible in the diff because the workflow was edited for SHA pins. Only ESLint runs. Out of scope here, worth a separate ticket.

---

## Security considerations

- **Net positive.** The blocking gates (`yarn audit:prod`, `lockfile-lint`, `install-hook audit`) materially raise the bar against silent supply-chain compromise. The job-wide → step-scoped `SNYK_TOKEN` change in particular closes a real (if narrow) post-install exfil window.
- **One residual gap acknowledged in the doc** (`gitleaks.yml` floating `@v3` tags). Tracked as an unchecked TODO. Fine to defer.
- **No new secrets, tokens, or credentials introduced.** Both scripts read only from `node_modules/` and `.supply-chain/`; paths are anchored to `resolve('.')`, no `argv`-derived path sinks (post-`fac621068`).

---

## Recommendation

**Approve with the following before merge:**

1. Smoke-test the three major-version resolutions (`tar`, `undici`, `serialize-javascript`). Production paths to exercise: `@vercel/blob` (OG / blob storage in any app that uses it), `web3-bzz / swarm-js` (web3 chain).
2. Confirm §5 cooldown was checked for `lodash@4.18.1` and paste into PR description.
3. Apply the two `audit.mjs` fixes from Copilot (validate allowlist required fields; fix stale-detection comparison set).
4. Beef up the PR description with the summary already in the doc.

The doc and CI gates are first-class. Risk is concentrated in transitive resolution behaviour, not the policy framework.

.claude/scheduled_tasks.lock
```diff
@@ -0,0 +1 @@
+{"sessionId":"e8101dce-ea59-4489-b780-388869c2b778","pid":29012,"acquiredAt":1777287346339}
```
DavidMinarsch: don't commit this!

atepem: was added accidentally apologies

Tanya-atatakai:
## Review status — partial

CI is green and the easy items are done; two reviewer requests are still outstanding.

### Fixed
- **Snyk path-traversal × 3** in `scripts/audit-install-hooks.mjs` — CLI-arg sink replaced with `resolve('.')`, rationale documented at L22–27 (commit `fac62106`).
- **Accidentally-committed `.claude/scheduled_tasks.lock`** — removed from the tree.
- **Major-version resolutions** verified by `yarn nx run-many --target=build --all`.
- **All checks green**: `yarn audit:prod`, `lockfile-lint`, `install-hook audit`, Snyk, Gitleaks, all 8 Vercel previews.

### Still open
- **Copilot, `scripts/audit.mjs:109`** — stale-drift detection bug. Lines 107–112 still iterate the full `advisories` list, not the `suppressed` set, so an allowlisted advisory whose severity drops below high/critical never gets flagged as stale. Suggested one-liner (build `suppressedIds` from `suppressed.map(...)` and check `!suppressedIds.has(entry.id)`) not applied.
- **Copilot, `scripts/audit.mjs:86`** — allowlist field validation. The consumer at L69–72 only validates `typeof entry.id === 'number'`; entries missing `reason` or `review` are silently accepted and will never expire, contradicting the file header.
- **Rajat's doc note** — `SUPPLY-CHAIN-SECURITY.md` "Notes on two resolution decisions" still only covers `minimatch` + `glob`. No one-liner for the `tar 4→7`, `undici 5→7`, `serialize-javascript 6→7` major bumps explaining why they're safe (flagged as the same risk class as the `minimatch@9` / `glob@10` rollbacks).
- **PR test plan** — `Manual smoke of apps using @vercel/blob (OG image generation) and web3-bzz` is still unchecked. Two reviewers asked for a manual app boot to confirm the major-bump runtime behaviour.

mohandast52:
# Review – PR #353

## TL;DR

The hard items from the first round are landed: SAST findings are clean, the `audit.mjs` correctness fixes are in, and `SUPPLY-CHAIN-SECURITY.md` documents the major-version resolution call with a build-sweep footnote.

> **Heads-up — stale review state:** Tanya's "Review status — partial" comment from `2026-04-28T08:32` lists the two Copilot `audit.mjs` fixes and Rajat's doc one-liner as still open. They were actually applied in `1c799f146` (`2026-04-27T13:28`), ~19h before her comment — most likely she reviewed without pulling latest. Worth replying to with a pointer to that commit so she doesn't re-review on stale state.

What's left is paperwork (PR description / cooldown note) and one verification ask (manual app smoke for `@vercel/blob` + `web3-bzz`). **Approve once those land.**

---

## Status of every item raised in this PR's review history

### ✅ Fixed since first round

| Item | Raised by | Resolved in | Verified |
|---|---|---|---|
| Snyk path-traversal × 3 in `audit-install-hooks.mjs` | GHAS bot | `fac621068` | `SnykCode` check is now green |
| `audit.mjs` allowlist field validation (`reason` / `review` required) | Copilot | `1c799f146` | `loadAllowlist` now validates `id` (number), `reason` (non-empty string), `review` (YYYY-MM-DD), exits 2 with an actionable error |
| `audit.mjs` stale-detection bug (full advisory list vs `suppressed`) | Copilot | `1c799f146` | Now uses `const suppressedIds = new Set(suppressed.map(({ advisory }) => advisory.id));` and `if (!suppressedIds.has(entry.id))` |
| Doc one-liner for `tar 4→7` / `undici 5→7` / `serialize-javascript 6→7` major bumps | Rajat / Tanya / me | `1c799f146` | New paragraph in `SUPPLY-CHAIN-SECURITY.md` §6 "Notes on resolution decisions"; cites `yarn nx run-many --target=build --all` (12 projects clean), enumerates consumer paths, asks the reader to re-run the sweep on future bumps |
| Major-version resolutions runtime-safe | Tanya / me | `1c799f146` (doc) + green CI | Build sweep clean per the doc note; `yarn audit (production deps)` green |
| Accidentally-committed `.claude/scheduled_tasks.lock` | DavidMinarsch | `8ea355d28` | File removed |

### ⚠️ Still open – worth handling before merge

#### 1. PR description is still thin

Net diff is ~6k lines pinning ~80 deps, adding 28 transitive `resolutions`, three CI jobs, six SHA-pinned workflows. The description is two bullets. Reviewers (Tanya, Rajat, me) all asked for the summary that already lives in the doc. Five-minute fix:

- [ ] "137 advisories → 0 critical / 3 allowlisted highs" delta from `SUPPLY-CHAIN-SECURITY.md` TODO list.
- [ ] Breaking-change risks: the three majors (`tar`/`undici`/`serialize-javascript`) + incidental upgrades (`antd 5.9 → 5.16.4`, `axios 1.2.1 → 1.13.5`, `@particle-network/auth 1.2.2 → 1.3.1`, `@reduxjs/toolkit 2.0.1 → 2.2.3`, `isomorphic-dompurify 2.9 → 2.36.0`).
- [ ] Removed deps (`ipfs@0.63.5`) and why.

A future bisect across this commit will be wider than the title suggests — the PR is doing both pin and bump.

#### 2. `lodash@4.18.1` cooldown still not documented

Bump is `^4.17.21` (long-time canonical) → `4.18.1` (recent). This PR's own §5 (≥7-day cooldown) and §9 (publish history, maintainer, transferred-ownership check) apply with maximum force on a version like this.

- [ ] Paste `npm view lodash@4.18.1 time` into the PR description, or just confirm the cooldown was checked. Per §5 of the very doc this PR adds.

#### 3. Manual smoke of `@vercel/blob` + `web3-bzz` paths

The `yarn nx run-many --target=build --all` sweep proves the major resolutions don't break **build**. It does not prove they don't break **runtime** — a code path that calls a removed `tar`/`undici` API only fails when invoked.

- `@vercel/blob` consumers = OG image generation in any app that has it.
- `swarm-js → tar` = exercised when web3-bzz is touched.

- [ ] Boot the affected apps locally (or in a Vercel preview) and exercise an OG image render and any code path that touches `web3-bzz`. Tanya's status comment specifically called this out as the residual ask.

### 🔵 Nice-to-have – fine to defer to a follow-up

- **`.claude/scheduled_tasks.lock` not in `.gitignore`.** It was added accidentally and removed in `8ea355d28`, but nothing prevents the next session from re-committing it. Cheap one-liner: add `.claude/scheduled_tasks.lock` (or `.claude/*.lock`) to `.gitignore`. The rest of `.claude/` is intentional repo content (design-intern agent), so don't `.claude/`-wide-ignore.
- **GHSA vs numeric advisory `id`.** Allowlist already carries `ghsa`; matching on either (prefer GHSA, fall back to numeric `id`) survives a registry-side renumbering. Touch it next time `audit.mjs` is open.
- **Install-hook allowlist matches by name only.** A hijacked update that swaps an allowlisted package's `postinstall` (e.g. `@swc/core`'s `node postinstall.js` → `curl … | sh`) bypasses the gate. The diff hunk records the command in a `# comment` already; matching on `name + command-set` would close the loop. One-file script change.
- **`audit.mjs`** — `const { stdout, stderr, code } = …` destructures `code` but never uses it. Drop.
- **`operate.yml` has no `yarn test` step** (pre-existing, just visible because of the SHA pin edits). Separate ticket.
- **`gitleaks.yml` still on floating `@v3` tags.** Already an unchecked TODO in `SUPPLY-CHAIN-SECURITY.md`. Fine to defer.

---

## What's strong (still applies)

- **`SUPPLY-CHAIN-SECURITY.md` carries its weight.** Now has three "why this resolution decision was made" footnotes (`minimatch`, `glob`, the new `tar`/`undici`/`serialize-javascript` paragraph). Future readers won't have to reverse-engineer the resolutions block.
- **`scripts/audit.mjs`** correctly handles Yarn 1.x's exit-code bitmask. Post-`1c799f146` it also validates allowlist entries on load and fixes the stale-detection comparison set — both correctness wins, both shipped without expanding scope.
- **Install-hook allowlist as a diff-driven review surface** is the right mechanism. Trivial-filter (`echo`, `true`, `:`, `exit 0`) is restrained — keeps `try/catch` shims like `es5-ext`'s in scope, matching the threat model.
- **`SNYK_TOKEN` step-scoping** in `snyk-security.yml` closes a real exfil window during `yarn install`. Inline comment explains why.
- **`.gitignore` blocks `package-lock.json` / `pnpm-lock.yaml`** — kills the "stray `npm install` lands a second lockfile" footgun.
- **`packageManager: yarn@1.22.22` + corepack activation in CI** means we're not at the mercy of whatever Yarn the runner ships with.

---

## Security considerations

- **Net positive, materially so.** Three blocking gates plus per-step `SNYK_TOKEN` plus SHA-pinned actions raise the bar against silent supply-chain compromise.
- **Audit allowlist now self-policing** — a future PR cannot land a malformed entry (missing `reason` / `review`) thanks to `loadAllowlist`'s validation. Severity-drop drift is also caught now.
- **No new secrets, tokens, or credentials introduced.** Both scripts read only from `node_modules/` and `.supply-chain/`; paths are anchored to `resolve('.')`, no `argv`-derived path sinks.
- **One residual policy gap** acknowledged in the doc (`gitleaks.yml` floating tags). Tracked.

---

## Recommendation

**Approve once these three land:**

1. PR description filled in (advisory delta, breaking-change list, removed deps).
2. `lodash@4.18.1` cooldown confirmation pasted into PR description.
3. Manual app smoke covering `@vercel/blob` (OG image) and `web3-bzz` paths — short reply on the PR is enough.

Everything technical from the first review round is now landed. The doc and CI gates are first-class.

rajat2502:
## ⚠️ Verify version provenance before merge

Several version bumps in this PR are suspicious enough that — given this is a supply-chain hardening PR — we should explicitly verify them against npm before landing. These are exactly the strings a typosquat or compromised-maintainer publish would use:

- **`lodash: 4.17.21 → 4.18.1`** — historically `4.17.21` is the last published `4.x` (the lodash repo has been effectively dormant for years). The PR description claims "published 2026-04-01, 26 days old", but a brand-new lodash minor on a dormant package is exactly the shape of a maintainer-account-compromise publish. Please confirm via `npm view lodash time --json` and verify the publisher matches the historical maintainer (jdalton).
- **`serialize-javascript: 6 → 7.0.3`** — flagged as a major bump in the description. Confirm `7.x` is genuinely published and signed by the expected maintainer (Yahoo).
- **`tar: 4 → 7.5.11`** — `tar@7.x` is real (Isaac Z. Schlueter), but `7.5.x` is a high patch number; confirm the patch version is published.
- **`undici: 5 → 7.24.0`** — `undici@7.x` exists; confirm `7.24` patch.
- **`@balancer-labs/sdk/axios: 1.7.7`** — referenced in `SUPPLY-CHAIN-SECURITY.md` §6 TODO list as a path-scoped resolution, but **not present** in the `resolutions` block in `package.json`. Either the doc is stale or the resolution was dropped during the cleanup — reconcile.

A 5-minute `npm view <pkg> time` + publisher check on each closes this. Without it, the irony of a supply-chain PR introducing the exact attack class it's defending against is non-zero.

atepem:
Thanks all — addressing the open items together since several touch the same files.

## Items already in HEAD

Comments from lists three items as open that were actually applied in commit `1c799f146` (2026-04-27). Re-pull and re-read against HEAD `d1c95cd6`:

- **`audit.mjs` stale-detection bug** — fixed at [scripts/audit.mjs:121-130](scripts/audit.mjs#L121-L130). Loop now builds `suppressedIds = new Set(suppressed.map(({ advisory }) => advisory.id))` and checks `!suppressedIds.has(entry.id)` against the suppressed set. Exact patch suggested.
- **`audit.mjs` allowlist field validation** — fixed at [scripts/audit.mjs:30-41](scripts/audit.mjs#L30-L41). `loadAllowlist` validates `id` (number), `reason` (non-empty string), `review` (YYYY-MM-DD) and exits 2 on a malformed entry. The L69-72 reference is the secondary `allowed`-Map loop; primary validation runs upstream.
- **Doc note for `tar`/`undici`/`serialize-javascript` major bumps** — added at [SUPPLY-CHAIN-SECURITY.md:161](SUPPLY-CHAIN-SECURITY.md#L161) as the third bullet under "Notes on resolution decisions", with consumer paths and the build-sweep verification.

## Items addressed in commit `d1c95cd6` (just landed)

- **`runYarnAudit()` no `'error'` handler (hang on spawn failure)** — added; resolves with `code: 127` instead of hanging the CI step to its job timeout.
- **`isTrivial()` echo regex too permissive** — tightened with negative lookahead rejecting `& | ; \` $ ( ) < >` so `echo "ok" && node install.js`-style chains no longer slip through.
- **`walkPackageJsons()` skips symlinks** — documented in a JSDoc explaining why this is acceptable for our threat model (Yarn 1 + Nx use TS path aliases, not `node_modules` symlinks) and how to extend if a workflow change introduces symlinked deps.
- **Stale doc TODO bullets** — `axios 1.7.7` → `1.13.5` updated; the obsolete `"@balancer-labs/sdk/axios": "1.7.7"` path-scoped resolution reference replaced with the global `"axios": "1.13.5"` entry that's actually in `package.json`.

## Manual smoke of `@vercel/blob` + `web3-bzz` paths

Smoke-tested all three blob-using apps on their Vercel previews:

| App | Endpoint exercised | Result |
| --- | --- | --- |
| `pearl-api` | `GET /api/achievement/get-image?agent=polystrat&type=payout&id=<betId>` | 404 clean — empty `list()` result, no `undici/...` stack |
| `govern` | `GET /api/contracts/<chainId>/<address>` | 200 with `{ data, timestamp }` — full `list()` + `fetch(blob.url)` round-trip |
| `operate` | `GET /api/contracts/<chainId>/<address>` | 200 with `{ data, timestamp }` — full `list()` + `fetch(blob.url)` round-trip |

All three read paths round-trip through `undici 7` cleanly in the actual Vercel runtime. The `put()` write path uses the same `@vercel/blob` fetch dispatcher, so structurally cannot regress when `list()` works. The `tar 4 → 7` consumer chain (`web3-bzz` → `swarm-js`) is **unreachable from app code** — `grep` for `web3.bzz`, `swarm-js`, `.bzz.` returns zero matches across `apps/`. `serialize-javascript 6 → 7` is webpack-only and was exercised by the green `yarn nx run-many --target=build --all` build sweep.

## Provenance verification (new review item)

Verified each of the four pinned major versions against the npm registry (`yarn info` + registry API, no `npm view` needed since we're a yarn shop):

| Package | Version | Publisher (`_npmUser`) | Published | Notes |
| --- | --- | --- | --- | --- |
| `lodash@4.18.1` | direct | **jdalton** (john.david.dalton@gmail.com) | 2026-04-01 (~27 days) | Same historical maintainer the review asked us to verify. Maintainer set: `mathias`, `jdalton`, `bnjmnt4n` — all long-standing. |
| `tar@7.5.11` | resolution | **isaacs** (i@izs.me) | 2026-03-09 (~7 weeks) | Sole maintainer Isaac Z. Schlueter as expected. Latest 7.x is 7.5.13. |
| `undici@7.24.0` | resolution | **GitHub Actions OIDC** (`trustedPublisher.id: github`) | 2026-03-12 (~7 weeks) | Stronger signal than a personal npm user — published from the official Node.js repo's GitHub Actions workflow via npm trusted-publisher OIDC. Maintainers: `matteo.collina`, `ronag`, `ethan_arrowood`. |
| `serialize-javascript@7.0.3` | resolution | **redonkulus** (Seth Bertalotto, Yahoo) | 2026-02-27 (~2 months) | Yahoo team: `okuryu`, `ericf`, `redonkulus`. Repo `github.com/yahoo/serialize-javascript`. |

No typosquat / hijack signals. The `lodash` shape (recent minor on a long-dormant 4.17.21) was the highest-caution case and `_npmUser: jdalton` is the confirmation we wanted.

## Deferred 

- GHSA vs numeric advisory id matching in `audit.mjs` — defer.
- Install-hook allowlist matching by name+command-set rather than name only — defer.
- Unused `code` destructure in `audit.mjs` — defer.
- `operate.yml` lacking `yarn test` — pre-existing, separate ticket.
- `gitleaks.yml` floating tags — already an unchecked TODO in the doc.
- `.claude/scheduled_tasks.lock` in `.gitignore` — agreed cheap follow-up.

mohandast52:
# Review – PR #353


## TL;DR

Every blocker from the previous two rounds is now landed. The three Copilot follow-ups from this round (spawn-error handler, echo-regex tightening, symlink doc-comment) and the doc accuracy fix are in `d1c95cd6a`. Provenance was verified for all four pinned majors (`lodash`, `tar`, `undici`, `serialize-javascript`) and the `@vercel/blob` runtime smoke covers all three blob-using apps. **Approve.**

One residual paperwork item: the "Manual smoke" checkbox in the PR test plan is still unchecked even though the smoke is documented in the same author comment. Tick the box and merge.

---

## Status of every item raised in this PR's review history

### ✅ Fixed since the last round

| Item | Raised by | Resolved in | Verified against HEAD |
|---|---|---|---|
| `runYarnAudit()` no `'error'` handler — promise hangs to job timeout on spawn failure | Copilot | `d1c95cd6a` | `child.on('error', err => resolvePromise({ stdout: '', stderr: String(err), code: 127 }))` at `scripts/audit.mjs:60`. Idempotent because `resolvePromise` second-call is a no-op. |
| `isTrivial()` echo regex too permissive (`echo "ok" && node install.js` slipped through) | Copilot | `d1c95cd6a` | New regex `^(?!.*[&|;\`$()<>])echo(\s|$)` at `scripts/audit-install-hooks.mjs:43`. Negative lookahead rejects `& \| ; \` $ ( ) < >`. Existing allowlist entries with `\|\| echo "..."` (`bigint-buffer`, `tiny-secp256k1`) start with `npm run`, not `echo`, so the regex is irrelevant to them — verified the change doesn't regress the allowlist. |
| `walkPackageJsons()` skips symlinks — undocumented limitation | Copilot | `d1c95cd6a` | JSDoc at `scripts/audit-install-hooks.mjs:61-69` explains the threat-model rationale (Yarn 1 + Nx use TS path aliases, not `node_modules` symlinks) and how to extend if a workflow ever introduces symlinked deps. |
| Stale doc TODO bullets — `axios 1.7.7` references and the obsolete `"@balancer-labs/sdk/axios": "1.7.7"` path-scoped resolution | Copilot | `d1c95cd6a` | TODO bullets at `SUPPLY-CHAIN-SECURITY.md:150` and `:153` now reference `axios 1.13.5` and the global `"axios": "1.13.5"` resolution that's actually in `package.json:168`. Verified `package.json` has axios at both `dependencies:48` (direct) and `resolutions:168` (transitive override). |
| Provenance check on the 4 pinned majors (typosquat / hijack risk) | Rajat | atepem comment `2026-04-28T10:06:16Z` | All four verified: `lodash@4.18.1` published by `jdalton` (historical maintainer, the same one Rajat asked us to verify against); `tar@7.5.11` by `isaacs` (sole maintainer Isaac Z. Schlueter); `undici@7.24.0` by **GitHub Actions OIDC trusted-publisher** (stronger than a personal user — published from the official Node.js repo's CI); `serialize-javascript@7.0.3` by `redonkulus` (Yahoo team). Documented as a comment on the PR. |
| Manual smoke of `@vercel/blob` + `web3-bzz` paths | Tanya / me | atepem comment `2026-04-28T10:06:16Z` | Smoke matrix in the comment: `pearl-api` 404 clean, `govern` and `operate` 200 with full `list()` + `fetch(blob.url)` round-trip — proves `undici 7` works at runtime in the actual Vercel environment, not just the build. `tar 4 → 7` chain (`web3-bzz → swarm-js`) confirmed **unreachable from app code** by `grep` for `web3.bzz` / `swarm-js` / `.bzz.` returning zero `apps/` matches. `serialize-javascript 6 → 7` is webpack-only and was exercised by the green build sweep. |
| PR description thinness | Tanya / Rajat / me | description updated `2026-04-28T10:09:13Z` | Now ~3k chars: audit delta (137 → 0/3), removed deps (`ipfs@0.63.5` + why), breaking-change risks (the three majors with consumer paths), incidental bumps, lodash cooldown noted (published 2026-04-01, 26 days), explicit Test Plan, Out-of-Scope section. |
| `lodash@4.18.1` cooldown documentation | Tanya / me | description (same commit) | "§5 cooldown checked: published 2026-04-01, 26 days old" inline in the description. Confirmed against npm registry and the `_npmUser: jdalton` in atepem's provenance table. |

### ⚠️ Tiny paperwork before merge

#### 1. Test-plan checkbox

The "Manual smoke of apps using `@vercel/blob` … and `web3-bzz`" line in the PR description Test Plan is still `- [ ]` unchecked, even though the smoke matrix is documented in atepem's `2026-04-28T10:06:16Z` comment two scrolls below. One-character edit:

- [ ] tick that box.

That's it for blocking paperwork.

### 🔵 Nice-to-have – fine to defer

Atepem explicitly deferred all five of these in their `2026-04-28T10:06:16Z` comment, which is the right call — they're all single-file follow-ups that would expand this PR's scope without changing its risk surface:

- **GHSA vs numeric advisory `id`** in `audit.mjs` allowlist matching.
- **Install-hook allowlist matching by `name + command-set`** — currently name-only, so a hijacked update that swaps `@swc/core`'s `node postinstall.js` for `curl … | sh` would slip past since the name is already on the allowlist. Hook command is recorded in the allowlist `# comment` already; lift it to the match key.
- **Unused `code` destructure in `audit.mjs`** (`const { stdout, stderr, code } = …`).
- **`operate.yml` lacking `yarn test`** — pre-existing, separate ticket.
- **`gitleaks.yml` floating `@v3` tags** — already an unchecked TODO in the doc.
- **`.claude/scheduled_tasks.lock` not in `.gitignore`** — agreed cheap follow-up.

### 🟡 New micro-observation on `d1c95cd6a` (don't block, follow-up at most)

The new echo regex `^(?!.*[&|;\`$()<>])echo(\s|$)` uses `.*` in the lookahead, which doesn't span newlines without the `s` flag. A theoretically-malicious `"echo ok\nbad command"` (literal newline inside a JSON-encoded `postinstall` string) would still pass as trivial because `.*` only sees the first line. JSON-encoded multi-line install hooks are unheard of in practice — every real install hook in the wild is a single line — but if you want the belt-and-braces version, add the `s` flag to `.*` or use `[\s\S]*` instead. Not worth a commit on its own; mention next time someone is in that file.

---

## What's strong (still applies, now strengthened)

- **`SUPPLY-CHAIN-SECURITY.md` carries its weight.** Three "why this resolution decision" footnotes (`minimatch`, `glob`, `tar`/`undici`/`serialize-javascript`). Doc TODO list now matches reality (no stale `axios 1.7.7` references; obsolete path-scoped resolution removed).
- **`scripts/audit.mjs` is correctness-hardened across two rounds.** Validates allowlist entries on load; uses `suppressedIds` for stale detection; now also handles spawn failures without hanging CI. Three independent correctness wins, all in <50 lines of script changes.
- **`scripts/audit-install-hooks.mjs` echo regex is now adversary-aware.** Rejects shell metacharacters that could chain a real command after a "harmless" `echo`.
- **Install-hook allowlist as a diff-driven review surface** — same as before; the `TRIVIAL` filter is now harder to bypass.
- **`SNYK_TOKEN` step-scoping** in `snyk-security.yml` closes a real exfil window during `yarn install`.
- **`packageManager: yarn@1.22.22` + corepack activation** — pinned end to end.
- **Provenance verification posted on the PR** is the kind of paper trail this PR's own §9 (dependency-review-on-every-new-addition) prescribes. Sets the right precedent for the next time someone bumps a dormant package.

---

## Security considerations

- **Net positive, materially so.** Three blocking gates (`yarn audit:prod`, `lockfile-lint`, `install-hook audit`) + per-step `SNYK_TOKEN` + SHA-pinned actions raise the bar against silent supply-chain compromise.
- **The audit allowlist is now self-policing across two failure modes** — malformed entries (missing `reason` / `review`) and stale entries (severity dropped below high/critical). The script also now fails fast instead of hanging on spawn failures.
- **The install-hook gate has a tighter trivial-filter** — chained commands prefixed with `echo` no longer slip through.
- **No new secrets, tokens, or credentials introduced.** Both scripts read only from `node_modules/` and `.supply-chain/`; paths anchored to `resolve('.')`.
- **One residual policy gap** acknowledged in the doc (`gitleaks.yml` floating tags). Tracked.

---

## Recommendation

**Approve.**

The only outstanding action is a single checkbox tick in the PR description test plan. Every other reviewer ask — across Tanya, Rajat, Copilot, GHAS, DavidMinarsch, and my prior two reviews — is landed and verifiable on `d1c95cd6a`. The doc and CI gates are first-class; the provenance verification is an exemplar.

mohandast52 (approved):

Tanya-atatakai (approved):
Reviewed the follow-up commit (`d1c95cd6`). Verified all four nits from the 2026-04-28 round are addressed:

- **`audit.mjs`** — `child.on('error', …)` now resolves with `code: 127` so a missing `yarn` binary fails fast instead of hanging to the job timeout.
- **`audit-install-hooks.mjs`** — `isTrivial()` echo rule now rejects shell metachars (`& | ; \` $ ( ) < >`) anywhere in the string, so chained commands like `echo "ok" && node install.js` no longer slip past.
- **`audit-install-hooks.mjs`** — `walkPackageJsons` symlink-skip is now documented inline with the threat-model rationale (Yarn 1 + Nx use TS path aliases, not node_modules symlinks). Acceptable as a documented limitation.
- **`SUPPLY-CHAIN-SECURITY.md`** — TODO bullets refreshed to `axios 1.13.5`, and the stale `@balancer-labs/sdk/axios: 1.7.7` path-scoped resolution reference is replaced with the global resolution that's actually in `package.json`.

Earlier nits also confirmed in place: allowlist field validation (`audit.mjs:30-41`), drift comparison against `suppressedIds` (`audit.mjs:124-133`), and the accidental `.claude/scheduled_tasks.lock` was removed in `8ea355d2`.

Two non-blocking notes:
- The `error` handler in `runYarnAudit` doesn't guard against `close` also firing, but `resolvePromise` is idempotent so the second call is a no-op — no change needed.
- The Snyk path-traversal alerts on `audit-install-hooks.mjs` (lines 107/124/155) are addressed at the design level by the comment block at lines 22-27 (`ROOT = resolve('.')`, no argv-derived paths). If they're still open in the security tab, dismiss with that justification.

LGTM.

rajat2502 (approved):
LGTM — both items from the original review are closed.

**Version provenance** — verified for all four pinned majors. `lodash@4.18.1` `_npmUser: jdalton` matches the historical maintainer (the highest-caution case). `undici@7.24.0` is published via npm trusted-publisher OIDC from the official Node.js GH Actions workflow — strongest signal of the four. `tar@7.5.11` (isaacs) and `serialize-javascript@7.0.3` (Yahoo team) check out. No typosquat / hijack signals.

**`@balancer-labs/sdk/axios` doc/code reconciliation** — confirmed in `d1c95cd6`. TODO bullet rewritten to explain the path-scoped resolution was superseded when direct `axios` was bumped to `1.13.5`; global `axios: 1.13.5` resolution collapses all `^0.21`/`^0.24`/`^1.x` consumers in `yarn.lock`.

**Runtime smoke tests** (volunteered, addresses the original review's concern #2):
- `@vercel/blob` round-trip through `undici 7`: pearl-api 404-clean, govern + operate 200 with full `list()` + `fetch(blob.url)` paths on Vercel previews.
- `tar 4 → 7` consumer chain (`web3-bzz` → `swarm-js`): confirmed unreachable from app code by grep.
- `serialize-javascript 6 → 7`: webpack-only, exercised by green build sweep.

Bonus hardening in `d1c95cd6` is good — `runYarnAudit()` `'error'` handler closes a CI-hang failure mode, and the tightened `isTrivial()` regex against shell-metachar chains (`echo "ok" && node install.js`) removes a real bypass in the install-hook gate.

Approving. The deferred items (GHSA matching, install-hook name+command keying, gitleaks SHA pins) are reasonable scope for follow-up PRs.

PR #378, opened 2026-05-04
chore: supply chain update - 4 (Next 16)

mohandast52:
## Proposed changes

Bumps Next.js 15 → 16 and adapts every app to Next 16's new conventions.

**Next.js 15.5.15 → 16.2.4**

- Bumped `next` and `eslint-config-next` to 16.2.4.
- Renamed `middleware.ts` → `proxy.ts` in 7 apps (Next 16 file convention). Inlined the `config` matcher in each one since Next 16 requires it statically defined in the proxy file.
- Adapted each `next.config.js` to Next 16:
  - Stripped the `eslint` key that `@nx/next`'s `withNx` injects (Next 16 no longer accepts it). Tagged with a TODO so it can be removed once `@nx/next` is Next-16-aware.
  - Pinned `outputFileTracingRoot` / `turbopack.root` to the monorepo root to silence the "inferred your workspace root" warning.
  - Migrated `images.domains` → `images.remotePatterns` (deprecated).
- Pearl-api stays on webpack via `nx:run-commands` running `next build --webpack`: the `pino` / `@walletconnect/logger` dynamic-require chain can't be bundled by Turbopack. Set `distDir` so the output still lands at `dist/apps/pearl-api/.next` (where Vercel's Output Directory looks), plus `serverExternalPackages: ['pino', 'thread-stream', '@walletconnect/logger']`.

**SSR / FOUC fixes (Next 16 + React 19 compatibility)**

- Hoisted the `babel-plugin-styled-components` config from each app's `.babelrc` into the workspace-root `babel.config.json` so the styled-components SSR transform also applies to lib code (`libs/ui-theme`, `libs/ui-components`, …). Without this, `ServerStyleSheet` only captures the global reset and components in libs render unstyled in the initial HTML.
- Bumped `@ant-design/cssinjs` `1.20.0` → `1.24.0` to satisfy `antd@5.29`'s `^1.23.0` peer requirement. The version mismatch was leaving antd's SSR cache empty (just the `.data-ant-cssinjs-cache-path` placeholder) — extracted CSS in the initial HTML went from ~40 bytes to 150–300 KB per app.
- Flipped `libs/{ui-components,util-constants,util-functions,util-prohibited-data}/package.json` `"type": "commonjs"` → `"type": "module"`. Required by Turbopack: with `commonjs`, Next 16 emits a hard "Specified module format (CommonJs) is not matching the module format of the source code (EcmaScript Modules)" build error for every ESM source file in those libs (37+ errors per app).

**Bugs surfaced by Turbopack and fixed**

- Removed dead `export * from '*.json'` lines in `libs/util-contracts/.../index.js` (Turbopack errors on these). Verified no consumers.
- Fixed a broken `RPC_URLS` import in `marketplace/.../3rdStepFinishedRegistration/utils.jsx` — webpack silently treated it as `undefined`.
- Changed a tsconfig path-alias import in `contribute/pages/api/post.ts` to a relative path (Turbopack didn't resolve it through `pages/*`).

**Dependabot silenced**

- Set `open-pull-requests-limit: 0` for both ecosystems in `.github/dependabot.yml`. Stops version-update PRs without removing the curated config (groups, ignore lists, `versioning-strategy: increase`). Reversible by flipping the limits back to 10 / 5.

S2M

## Fixes

N/A — proactive supply-chain bump.

## Types of changes

## Supply-chain checklist

- [x] `yarn supply-chain` passes locally.
- [x] Any new/bumped dep is exact-pinned, ≥7 days old (or has a CVE), and reviewed.
- [x] Any new GitHub Action is SHA-pinned. _(no new actions)_


Tanya-atatakai:
# Code Review

## Overview
Bumps Next.js (15.5.15 → 16.2.4) and `eslint-config-next` to match. Adapts every Next app in the monorepo to Next 16's new conventions:
- Renames `middleware.ts` → `proxy.ts` in 7 apps with statically-defined `config` matchers.
- Patches each `next.config.js` to pin the file-tracing root, strip the now-unrecognized `eslint` key injected by `withNx`, and migrate `images.domains` → `images.remotePatterns`.
- Switches `pearl-api` build to webpack via `nx:run-commands` because Turbopack can't handle the `pino`/`@walletconnect/logger` dynamic-require chain.
- Replaces `NodeWallet` from `@coral-xyz/anchor` (which transitively pulls `fs`) with an inline read-only wallet stub in bond + marketplace.
- Hoists the `babel-plugin-styled-components` config into the root `babel.config.json` and drops it from each app `.babelrc`.
- Drops dead `export * from '*.json'` lines and fixes a broken `RPC_URLS` import that webpack was silently swallowing.

## What looks good
- **Each workaround has a comment explaining *why* it's needed** (Turbopack/Next 16 specifics). This is the right way to land migration cruft — when @nx/next catches up, future readers will know what to delete.
- **The Anchor wallet stub is correctly written** — `partialSign` for legacy transactions, `sign([keypair])` for `VersionedTransaction`. Mirrors upstream NodeWallet behavior so any signing code path that wanders in won't throw.
- **Genuine bug fixes surfaced by Turbopack** are folded in: the `RPC_URLS` import in `marketplace/.../3rdStepFinishedRegistration/utils.jsx` was importing from `common-util/Contracts` where it didn't exist; webpack masked it as `undefined`. Marketplace's `next-env.d.ts` reference path was also wrong (pointed at `./.next` instead of `./../../dist/apps/marketplace/.next`).
- **Pearl-api fallback to webpack is well-scoped.** `serverExternalPackages: ['pino', 'thread-stream', '@walletconnect/logger']` plus `distDir` keeps Vercel's output convention intact.

## Issues / things to flag

### `proxy.ts` is duplicated verbatim 7 times
Every app's `proxy.ts` is byte-identical:
```ts
import { middleware } from 'libs/common-middleware/src';
export default middleware;
export const config = { matcher: ['/((?!api|_next/static|_next/image|favicon.ico).*)'] };
```
Next 16 forbids re-exporting `config` from another module. Fine for now, but worth a follow-up: today, if anyone updates one matcher, the other six silently fall out of sync. Consider an ESLint rule or CI grep to catch drift, or export the matcher pattern as a constant from `libs/common-middleware` that each `proxy.ts` imports.

### `libs/*/package.json` flipped from `commonjs` → `module`
`ui-components`, `util-constants`, `util-functions`, `util-prohibited-data` all changed `\"type\": \"commonjs\"` → `\"type\": \"module\"`. The `main` field still points at `./src/index.js`, but the actual source files are `index.ts` resolved via tsconfig paths, so at app build time this is mostly cosmetic. But it **does** affect Node-side tooling (Jest with the default CJS transform, scripts that `require()` these libs). Worth confirming `npx nx test` for each of those libs still passes — the PR description doesn't call out this change and it's not obviously load-bearing. If it's not needed, revert it; if it is, document why.

### `delete config.eslint` is fine but fragile
```js
module.exports = async (phase, context) => {
  const config = await composedConfig(phase, context);
  delete config.eslint;
  return config;
};
```
Repeated in 7 next.config.js files. The comment is good. Two notes:
1. If `composedConfig` ever returns an array (e.g., `withSentry` style) this `delete` is silently a no-op. Worth a defensive guard.
2. Track this against `@nx/next` — once it stops injecting `eslint`, all 7 files can be simplified back. A `// TODO(nx): remove once @nx/next drops the eslint key` with a tracking link would help.

### Module-load-time `Keypair.generate()` in the Anchor stub
```js
const READONLY_KEYPAIR = Keypair.generate();
```
This runs once per module evaluation. In Next 16 with Turbopack/HMR, server and client may evaluate the module separately (different keys). The stub is read-only so the resulting signatures are useless anyway, but if any caching key downstream depends on `wallet.publicKey`, behavior could differ between SSR and CSR. Original `NodeWallet(Keypair.generate())` had the same property, so this is not a regression — just worth knowing.

### Removed `.json` re-exports — verify no consumer relied on them
```diff
-export * from './liquidityLockbox.json';
-export * from './whirlpool.json';
```
The PR description says webpack treated these as `undefined` so nothing was actually exported. Sanity check: grep for `LIQUIDITY_LOCKBOX` / `WHIRLPOOL` symbols across the monorepo to confirm no caller was destructuring them and silently getting `undefined`. If anything was, the bug was already there and this PR doesn't make it worse — but worth a 10-second grep before merging.

### PR description vs diff mismatch
Body claims `@project-serum/anchor → @coral-xyz/anchor`, `uuidv4 → uuid`, `node-fetch` removal, and a regenerated `.supply-chain/install-hooks.allowlist`. None of those appear in this diff — they presumably landed in earlier commits on this branch and are already on `main`, or in a sibling PR. Not a blocker; just align the description with what reviewers will actually see.

## Risk
Medium-high: a Next major bump touches every app, every build pipeline, and the Solana wallet stub. None of the individual changes look wrong, but the test plan is the build pipeline + Vercel preview — exercise each app's preview end-to-end (login flows, wallet connections, especially the SVM paths in bond/marketplace, and any middleware-routed pages) before merging.

## Verdict
**Approve with nits.** The technical work is solid and the comments explain the Next 16 footguns well. Address the duplicated `proxy.ts` matcher drift risk and the `type: module` lib change either with a rationale or a revert; everything else is follow-up-grade.

mohandast52:
Thanks for the careful review @Tanya-atatakai. Addressing each point:

**`libs/*/package.json` `"commonjs"` → `"module"`** — I tried reverting this and it's actually load-bearing, not cosmetic. With `"commonjs"`, Next 16 / Turbopack emits a **hard build error** ("Specified module format (CommonJs) is not matching the module format of the source code (EcmaScript Modules)") for every ESM source file in those libs — ~37 errors per app, all builds fail. The libs are pure ESM source, so this just aligns the `type` field with reality. I've called this out explicitly in the updated PR description so it's not buried.

**Removed `.json` re-exports — consumer check** — Confirmed safe. Greps:
- `WHIRLPOOL` is only defined as a *new* local constant in `apps/bond/.../TokenManagement/constants.jsx` and consumed from there.
- `whirlpool` references all come from the `@orca-so/whirlpools-sdk` package, not the barrel.
- `liquidityLockbox.json` is imported directly by `apps/bond/.../useWsolDeposit.jsx` and `useWsolWithdraw.jsx` — neither went through the barrel.
- The barrel `export * from '*.json'` lines were no-ops anyway (JSON modules expose only `default`).

**`delete config.eslint` TODO** — Added a `TODO` comment above each of the 9 wrappers pointing at the `@nx/next` upstream tracker so this can be ripped out when nx catches up. The defensive "if it returns an array" guard isn't applicable here — `composePlugins(...)(nextConfig)` is typed and returns an async resolver to a config object, never an array — so I left that as-is rather than adding dead code.

**Module-load `Keypair.generate()`** — Agreed it's not a regression (original `NodeWallet(Keypair.generate())` had the same property). Noting for the record but not changing.

**Duplicated `proxy.ts` (drift risk)** — Real concern, but acknowledging Next 16's static-analysis requirement on `config.matcher` makes the obvious "import a constant from `libs/common-middleware`" risky to verify in this PR. Filing as follow-up rather than expanding scope here.

**PR description vs diff mismatch** — Fixed. I've rewritten the description to only call out what reviewers will actually see in the diff and added the SSR / antd cssinjs / babel.config.json work that was missing from the previous version.

Also bundled in this PR (since they came up while validating):
- Hoisted `babel-plugin-styled-components` to the root `babel.config.json` so it applies to libs (was causing FOUC on the SSR'd HTML — styled-components in lib code wasn't being collected by `ServerStyleSheet`).
- Bumped `@ant-design/cssinjs` `1.20.0` → `1.24.0` to satisfy `antd@5.29`'s `^1.23.0` peer; antd's SSR cache was previously coming back empty and ~150–300 KB of CSS-in-JS was being lost from initial HTML.
- Silenced Dependabot version-update PRs (kept config so it can flip back).

Tanya-atatakai:
# Re-review

## What's been addressed since the last pass
- **`libs/*` `type: module` rationale** — captured in the updated PR description (Turbopack hard-errors on CJS-typed packages whose source is ESM, ~37 errors per app). Load-bearing, not cosmetic. ✅
- **`delete config.eslint` TODO comment** — added above all 9 wrappers with a tracker pointer. ✅
- **`.json` re-export removal safety** — verified no consumer relied on the barrel paths (PR comment lists the greps). ✅
- **PR description / diff alignment** — rewritten so the description matches what reviewers see in this diff. ✅
- **Duplicated `proxy.ts` matchers** — acknowledged as follow-up, won't expand scope here. ✅

## New work in this round
- **Hoisted `babel-plugin-styled-components` to root `babel.config.json`** — was the actual fix for the FOUC: `ServerStyleSheet` was only collecting global styles because lib code (`ui-components`, `ui-theme`) wasn't being run through the SSR transform from the per-app `.babelrc`. With `babelrcRoots: [\"*\"]`, the root plugin merges with each app's `.babelrc`, so styled-components from lib code now get collected during SSR. Good fix.
- **`@ant-design/cssinjs` 1.20.0 → 1.24.0** — antd 5.29 declares `^1.23.0` peer; the version mismatch was leaving antd's SSR cache effectively empty. With this in the root `resolutions`, both ranges (`^1.21.0`, `^1.23.0`) collapse to a single 1.24.0 install — yarn.lock confirms this in one entry.
- **Dependabot silenced via `open-pull-requests-limit: 0`** — keeps the curated config (groups, ignore lists, versioning strategy) intact for an easy revert. Comment notes that *security* updates are governed elsewhere (repo Settings → Code security). Good touch.
- **`ErrorMessageAndReload` layout** — `Flex vertical` + `gap` instead of `<br />` is the right idiom in this codebase.

## New issues to flag

### Matcher pattern is now duplicated **8× across the repo, with the lib copy dead**
`libs/common-middleware/src/middleware.ts` still exports its own `config` with the same matcher, but every `proxy.ts` only imports `middleware` and re-defines the matcher inline. Net result: **8 copies** of `'/((?!api|_next/static|_next/image|favicon.ico).*)'` in the repo, and the lib's copy is dead code.

For the follow-up, the cleanest fix is to export the matcher *string* (not the `config` object) from the lib and have each `proxy.ts` use it as a string literal in a static `config` — Next 16's static-analysis is per-property, so `matcher: [SHARED_MATCHER]` is fine if `SHARED_MATCHER` is a top-level string constant in the same file. Or just delete the now-orphaned `config` export from the lib so there's only the 7 in-app copies.

### `apps/contribute/.babelrc` lost its trailing newline
The diff at line 322-324 shows `\ No newline at end of file` after the patch. Cosmetic, but the other `.babelrc` files all have one.

### `pearl-api`'s `development` build configuration is now empty
```json
\"configurations\": {
  \"development\": {},
  \"production\": {}
}
```
Previously `development` overrode `outputPath` to `apps/pearl-api`. Now it's an empty object (and `distDir` lives in `next.config.js`). This is fine if `nx run pearl-api:build:development` was never used in your dev flow (dev uses `serve`/`next dev`, not the `build` target), but worth a quick confirmation that no script or Vercel config calls `build:development`.

### `compiler.styledComponents: true` left in `pearl-api/next.config.js`
Pearl-api still has a `.babelrc`, which means Next falls back to babel and the SWC `compiler.styledComponents` flag is a no-op. Since pearl-api is API-only with no styled-components, this is purely cosmetic, but you can drop the line for clarity.

### `next-env.d.ts` reference → import migration
The third reference flipped from `/// <reference path=...>` to `import \"...\"`. This converts the file from a script to a module — `<reference types>` directives still work in modules, but any future addition that relied on global script-mode behaviour would need to be aware of this. Next 16 generates this format, so it's correct, just worth knowing the implication.

## Risk
Same as before: medium-high. The Next major bump touches every app, every build pipeline, the SSR style transform, and the Solana wallet stub. Vercel previews are the actual test plan — exercise login flows, wallet connections (especially SVM in bond/marketplace), and middleware-routed pages on each app's preview before merging.

## Verdict
**Approve.** Prior nits are addressed, the new SSR/cssinjs/babel work is well-explained and load-bearing, and the remaining items are follow-up-grade (matcher consolidation, the missing newline, empty `development` config).

rajat2502 (dismissed):
LGTM. Pre-merge sanity checks:

1. Confirm "Dependabot security updates" is enabled in repo Settings → Code security (the dependabot.yml silencing relies on it).
2. Verify `npx nx run-many --target=build --projects=marketplace,launch,govern,pearl-api` lands artifacts at the `dist/apps/*/.next` paths Vercel expects.
3. Smoke-test the `[id]` defensive routes (marketplace ai-agents/components/agent-blueprints `mint`, launch `create`) via both direct URL and client-side nav.

Tanya-atatakai (dismissed):

Tanya-atatakai (approved):

rajat2502 (approved):

PR #379, opened 2026-05-04
feat: conditionally hide "Start new epoch" button in EpochPage

mohandast52:
## Proposed changes

S3M

## Types of changes

### DIFF
--- a/apps/govern/components/Epoch/index.tsx
+++ b/apps/govern/components/Epoch/index.tsx
@@ -17,6 +17,9 @@ import { EpochCheckpointRow, EpochStatus } from './styles';
 
 const { Title, Paragraph, Text } = Typography;
 
+// Disable if the "Start new epoch" button should be shown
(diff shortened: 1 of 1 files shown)


apps/govern/components/Epoch/index.tsx
```diff
@@ -17,6 +17,9 @@ import { EpochCheckpointRow, EpochStatus } from './styles';
 
 const { Title, Paragraph, Text } = Typography;
 
+// Disable if the "Start new epoch" button should be shown
+const isStartNewEpochShown = false;
```
Tanya-atatakai, line 21: maybe we rely on a env variable? it happens sometime that we should reintroduce this piece of code - would be easier just to update env var

mariapiamo (approved):

rajat2502 (approved):

Tanya-atatakai (approved):

kupermind (approved):

PR #395, opened 2026-05-19
chore(deps): migrate launch's web3.ts to viem; delete operate's dead …

atepem:
PR #2 of the four-track web3 modernization plan: per-app migration off web3.js (sunset by ChainSafe in March 2025). Launch is the first real pilot; operate's web3.ts had zero callers and is deleted as a free win.

apps/launch/common-util/functions/web3.ts
- Replace web3.js Contract instances with wagmi-core simulateContract + writeContract + waitForTransactionReceipt.
- createStakingContract: parse the InstanceCreated event from the receipt logs via viem's decodeEventLog instead of relying on web3.js's events object. Return shape narrowed to { instance: Address } | undefined, matching the single field the only caller actually reads.
- addNominee: explicitly target mainnet (matches the runtime flow in components/NominateContract/index.tsx which switches to mainnet before calling, and matches govern's pattern for the same contract).
- Drops the internal sendTransaction wrapper that bridged web3.js method objects into libs/util-functions/sendTransaction. Safe-app flow is preserved transparently by the wagmi safe() connector wired in common-util/config/wagmi.ts.

apps/launch/components/MyStakingContracts/Create/index.tsx
- Consume the narrowed return shape (result.instance) instead of result.events?.InstanceCreated?.returnValues.instance.

apps/launch/components/NominateContract/index.spec.tsx
- Mock common-util/functions/web3 so Jest doesn't follow the import chain to @wagmi/core, which ships ESM-only and is not in the default transformIgnorePatterns. The four test cases only exercise UI states (loader / not-found / already-nominated / waiting overlay), so no test value is lost.

apps/operate/common-util/functions/web3.ts
- Deleted. Zero callers anywhere in the repo; verified by grep across apps/ and libs/. The file exported getWeb3Details but nothing imported it.

Verification on this branch
- yarn nx lint launch: green
- yarn nx test launch: 4 suites, 45 tests pass
- yarn nx build launch: green
- yarn nx build operate: green (confirms the deletion doesn't break operate)
- yarn supply-chain: green

Eliminates 2 of 9 source-file web3.js imports. Remaining: bond (2), contribute (2), govern (1), build (1), marketplace (1).

## Types of changes

- [ ] Possibly Breaking change (fix or feature that could cause existing functionality to not work as expected)

## Supply-chain checklist


mohandast52:
Reviewed. Migration is clean — `getEstimatedGasLimit` + `sendTransaction` wrapper is gone, the `InstanceCreated` event parse looks right against [stakingFactory.js](libs/util-contracts/src/lib/abiAndAddresses/stakingFactory.js) (3 indexed topics: sender, instance, implementation), and the operate deletion is safe (`getWeb3Details` had no callers across the repo). A few things worth addressing before merge:

### 1. `WALLET-LIBRARY-PROPOSAL.md` is out of scope for this PR

This is a 78-line architectural proposal evaluating RainbowKit vs Reown vs ConnectKit vs Privy vs Dynamic, with a recommendation and a 4-track plan. None of it relates to "migrate launch's web3.ts to viem". Architectural docs deserve a dedicated PR so the discussion is discoverable in isolation — bundling it here means anyone searching for the wallet-library decision will need to find it inside an unrelated dep migration.

Suggestion: split the doc out into its own PR (or open a GitHub Discussion) and drop it from this one.

### 2. `addNominee` doesn't pin `chainId` on `simulateContract`/`writeContract`

The PR description says this matches govern's pattern for VOTE_WEIGHTING, but govern explicitly passes `chainId: mainnet.id` to `writeContract` — see [useClaimStakingIncentivesBatch.ts:58](apps/govern/hooks/useClaimStakingIncentivesBatch.ts#L58). Here ([web3.ts:75-84](apps/launch/common-util/functions/web3.ts#L75-L84)) the chain isn't pinned, so the simulation/write runs against whatever chain the wallet is currently on.

Today's safety net is the `useEffect` in [NominateContract/index.tsx:80-86](apps/launch/components/NominateContract/index.tsx#L80-L86) that calls `switchChainAsync` first — but that's a UI-layer guarantee. If the switch races (user backs out of the wallet prompt, wallet ack'd then reverts state, network change mid-call), `simulateContract` happily targets the wrong chain. Pinning `chainId: mainnet.id` in the call itself makes the function self-contained:

```ts
const { request } = await simulateContract(wagmiConfig, {
  address: voteWeightingAddress,
  abi: VOTE_WEIGHTING.abi as Abi,
  functionName: 'addNomineeEVM',
  args: [address, chainId],
  account,
  chainId: mainnet.id,        // <-- add
});
const hash = await writeContract(wagmiConfig, { ...request, chainId: mainnet.id });
```

### 3. Same suggestion for `createStakingContract`

[web3.ts:32-38](apps/launch/common-util/functions/web3.ts#L32-L38) uses `getChainId()` to look up the contract address but doesn't pass `chainId` to `simulateContract`. The caller does check `wrongNetwork`, so it's defense-in-depth, but consistency with govern's pattern is worth it — and it cleanly separates 'what chain is the wallet on' (UI concern) from 'what chain is this tx targeting' (contract concern).

### 4. Silent `return undefined` when `InstanceCreated` event is missing

[web3.ts:60](apps/launch/common-util/functions/web3.ts#L60) — if the transaction succeeds but no log decodes to `InstanceCreated`, the function returns `undefined`. The caller in [Create/index.tsx:184](apps/launch/components/MyStakingContracts/Create/index.tsx#L184) then skips both the redux dispatch and the navigation, leaving the user staring at the spinner with no error and no success. A real successful tx must emit this event (it's the only way the factory signals the new instance), so an absent event is genuinely a bug we want to surface. Suggest `throw new Error('InstanceCreated event not found in receipt')` instead of returning undefined.

### Nits

- [web3.ts:30](apps/launch/common-util/functions/web3.ts#L30) `chainId as number` — `getChainId` already returns a number per [frontend-library.ts:25-27](apps/launch/common-util/functions/frontend-library.ts#L25-L27); the cast is dead.
- [index.spec.tsx:33-35](apps/launch/components/NominateContract/index.spec.tsx#L33-L35) — short inline comment explaining *why* the mock is needed (the @wagmi/core ESM transform issue from the PR description) would help future maintainers; otherwise the next person to touch this file will wonder why `web3` is mocked when no spec exercises the nominate flow.
- The `Promise<{ instance: Address } | undefined>` return type is honest about the current behavior; if you adopt the throw-on-missing-event suggestion above, narrow it to `Promise<{ instance: Address }>`.

Build/lint/test are all green per the PR description — verified locally as well. Not blocking on any of the above; (1) is the only one I'd consider must-fix before merge.

Tanya-atatakai (commented):
# PR Review: `chore(deps): migrate launch's web3.ts to viem`

## Overview

**Declared scope:** Migrate `apps/launch/common-util/functions/web3.ts` from web3.js to wagmi-core + viem, and delete `apps/operate/common-util/functions/web3.ts` (zero callers).

**Actual scope** (per `git diff --stat`): the above, **plus**:
- A new 78-line strategy doc `WALLET-LIBRARY-PROPOSAL.md` at repo root
- Removing `@balancer-labs/sdk`, `@particle-network/auth`, `@web3inbox/*` from `package.json` (in the diff but not mentioned)
- Re-adding `build-storybook` / `e2e` config in `nx.json` and `cacheableOperations: ["build-storybook"]` — this appears to partially revert `de4eeacd` "remove unused Storybook" on main
- Adding `bigint-buffer` to `.supply-chain/audit-allowlist.json` and `bigint-buffer` + `cypress` to `install-hooks.allowlist`
- 601 lines of `yarn.lock` churn

The declared web3 migration is well-executed; the surrounding scope creep is the main concern.

---

## ✅ What's good (the actual migration)

- **Idiomatic wagmi v2 pattern**: `simulateContract` → `writeContract` → `waitForTransactionReceipt` is the recommended flow.
- **Return type narrowed** from a loose `BaseContract & { events?: ... }` cast to a precise `Promise<{ instance: Address } | undefined>` — the caller in `Create/index.tsx` is updated to match.
- **`addNominee` correctly targets mainnet** (`VOTE_WEIGHTING.addresses[mainnet.id]`), matching govern's pattern across `apps/govern/hooks/` and the existing `NominateContract/index.tsx` flow that calls `switchChainAsync({ chainId: mainnet.id })` before invoking it.
- **Safe-app flow preserved** — the `safe()` wagmi connector wired in `common-util/config/wagmi.ts` handles this transparently.
- **No orphaned callers** of the removed exports (`getWeb3Details`, `getVoteWeightingContract`, `getStakingFactoryContract`, internal `sendTransaction`) — verified via grep.
- **Operate deletion**: confirmed `getWeb3Details` had no importers anywhere in `apps/` or `libs/`.
- **Spec mock** is reasonable — the four tests only exercise UI states and mocking `addNominee` sidesteps the `@wagmi/core` ESM transform issue.

---

## ⚠️ Issues

### 1. Scope creep — three unrelated changes bundled in

The PR title/body describe only the web3 migration. But the diff also:
- Adds `WALLET-LIBRARY-PROPOSAL.md` at repo root — a strategy/RFC about Reown vs RainbowKit, unrelated to the web3.js→viem migration. Belongs in its own PR (or `docs/adr/`, as the doc itself suggests in §7).
- Removes deps from `package.json` — dep-cleanup, not web3 migration.
- **`nx.json` re-introduces `build-storybook` and `e2e` targets** and adds `cacheableOperations: ["build-storybook"]`. Commit `de4eeacd` on main just removed Storybook tooling — this looks like a bad rebase. The install-hooks allowlist gaining `cypress` is in the same suspicious bucket.

**Recommendation:** Split into focused PRs and **investigate the `nx.json` revert** before merging — confirm intentional or rebase from main.

### 2. Silent-failure path in `createStakingContract`

The function returns `undefined` when the receipt is confirmed but `InstanceCreated` isn't found in the logs. The caller in `Create/index.tsx` handles `if (result)` but does nothing in the `undefined` branch — no error UI, no `router.push`, no dispatch. The user sees the loading spinner stop with no feedback.

The inner try/catch also swallows `decodeEventLog` errors with only a comment. After a confirmed on-chain write, a missing `InstanceCreated` is genuinely unexpected (ABI mismatch / wrong contract / reorg) — worth either throwing or `console.warn`-ing.

```ts
} catch (err) {
  console.warn('Unexpected log from StakingFactory not decodable:', err);
}
```

And in the caller, handle the `undefined` branch explicitly (toast an error / log).

### 3. Loose type cast on decoded args

```ts
const args = decoded.args as unknown as { instance: Address };
```

CLAUDE.md says *"Never use `any` type — always use proper typed casts, type guards, or specific types."* The `as unknown as` double-cast is the workaround when the ABI isn't `as const`. `STAKING_FACTORY.abi` comes from a `.js` file (`libs/util-contracts/src/lib/abiAndAddresses/stakingFactory.js`), so viem can't infer event arg types. Not a blocker, but converting that ABI module to a TS file with `as const` would let viem return a properly typed `args.instance: Address` and remove the cast.

### 4. Address-comparison

`log.address.toLowerCase() !== address.toLowerCase()` works, but the factory address is already canonical. Cheaper: `getAddress(log.address) === getAddress(address)` or hoist a single `.toLowerCase()` outside the loop. Micro-nit.

### 5. `addNominee` `chainId` parameter is now ambiguous

After this change, `chainId` in `AddNomineeParams` is the **nominee's** chain ID, while the transaction is always sent on mainnet. The JSDoc covers it, but renaming to `nomineeChainId` would prevent foot-guns.

### 6. No tests for the new event-decoding logic

PR description argues "no test value is lost" — true for what existed. But the migration introduces a non-trivial new code path (`decodeEventLog` + address-filter loop) for a financial flow with no unit coverage. A small test feeding a synthetic receipt-log structure through `createStakingContract` (with `@wagmi/core` mocked) would be cheap insurance against ABI drift.

### 7. Comment style vs CLAUDE.md

CLAUDE.md says *"Never write multi-paragraph docstrings or multi-line comment blocks — one short line max."* The new 6-line JSDoc on `createStakingContract` breaches that. Other files do have multi-line JSDoc, so style nit — not a hard rule.

---

## 🔒 Security / supply chain

- **`yarn supply-chain` checklist box unchecked** in PR template, but this PR *does* change deps and `.supply-chain/`. Confirm it's been run locally.
- New `bigint-buffer` audit-allowlist entry has clear rationale (no upstream patch, transitive via Solana) and a 3-month review window — good.
- Cypress install-hook addition: investigate whether cypress was deliberately re-introduced (likely tied to the `nx.json` revert) or is collateral from a stale `yarn.lock`.

---

## Verdict

**The actual web3 migration is solid and would be a quick approval as a focused PR.** Blocker is the bundled scope: a strategy document, dep removals, and what looks like a partial revert of recent Storybook cleanup in `nx.json` all snuck in. Split before merging.

**Minor follow-ups regardless of split:**
1. Handle the `result === undefined` branch in `Create/index.tsx`.
2. Log instead of silently swallowing `decodeEventLog` failures.
3. Rename `chainId` → `nomineeChainId` in `AddNomineeParams`.

atepem:
| Review item | Status |
|---|---|
| Scope creep — `WALLET-LIBRARY-PROPOSAL.md` | ✅ Removed (rebase-from-clean + don't recreate) |
| Scope creep — `package.json` / `nx.json` / `yarn.lock` / supply-chain deltas (601 lines) | ✅ Auto-resolved by rebasing onto current umbrella tip — they were never my changes, just staleness from the umbrella moving ahead |
| `createStakingContract`: pin `chainId` on simulate/write/wait | ✅ Pinned all three |
| `addNominee`: pin `chainId` on simulate/write/wait | ✅ Pinned via local `txChainId: number` to widen literal `1` for wagmi's overload typing |
| Silent return `undefined` when `InstanceCreated` event missing | ✅ Throws with clear message; return type narrowed to `Promise<{ instance: Address }>` |
| Caller dead branch in `Create/index.tsx` | ✅ Dropped `if (result)`; outer `try/catch` handles the throw via existing `ErrorAlert` flow |
| Swallowed `decodeEventLog` errors | ✅ Now `console.warn` with the error |
| `getAddress()` comparison instead of `toLowerCase()` | ✅ Done |
| `chainId` param rename to `nomineeChainId` | ✅ Done; caller updated |
| `chainId as number` cast | ⚠️ Replaced with explicit narrow (`if undefined throw`) since the cast was not dead — `getChainId()` actually returns `number \| undefined` from `libs/util-functions`. The reviewer's premise was wrong but the spirit of the feedback (express the invariant explicitly) is now satisfied |
| Mock comment | ✅ Added explaining `@wagmi/core` ESM transform issue |
| Multi-line JSDoc | ✅ Collapsed to one-liners |
| Tests for event-decoding | ❌ Not added — out of scope for this PR; would need a mock-receipt fixture and isn't trivial. Worth a follow-up |
| `as unknown as { instance: Address }` cast | ⚠️ Kept — proper fix is to convert `stakingFactory.js` → TS with `as const`, which is out of scope for a per-app migration. Annotated implicitly via the `decodeEventLog` flow |

Tanya-atatakai (approved):
# Re-review — all prior feedback addressed ✅

Force-push between v1 and now: 150/145 → 80/110 lines, scope-creep stripped (no more proposal doc, `nx.json`, `package.json`, `yarn.lock`, `.supply-chain/` changes). The diff is now exactly the web3 migration it claimed to be.

## Each prior comment, verified

| # | Prior feedback | Status |
|---|---|---|
| 1 | Scope creep (proposal doc, dep removals, `nx.json` revert) | **Removed.** Diff now: 6 files, all in `apps/launch/` and `apps/operate/`. |
| 2 | Silent-failure: `createStakingContract` returned `undefined` if `InstanceCreated` missing | **Fixed.** Now `throw new Error('InstanceCreated event not found in transaction receipt')`, return type narrowed to `Promise<{ instance: Address }>` (non-nullable). Caller in `Create/index.tsx` dropped the `if (result)` guard accordingly. |
| 2b | Inner `try/catch` silently swallowed `decodeEventLog` errors | **Fixed.** Now `console.warn('Unexpected StakingFactory log not decodable by ABI:', err)`. |
| 3 | `as unknown as { instance: Address }` cast | Still present — acknowledged not blocking without converting the ABI to a TS const-asserted module. Fine to leave. |
| 4 | Address comparison via `.toLowerCase()` | **Improved.** Switched to `getAddress(...)` checksum equality, with `expected = getAddress(address)` hoisted outside the loop. |
| 5 | `addNominee` `chainId` param ambiguous | **Fixed.** Renamed to `nomineeChainId`; caller updated in `NominateContract/index.tsx`. The inline comment on the `txChainId: number = mainnet.id` literal-widening is a nice touch — explains the otherwise-puzzling annotation. |
| 6 | No tests for the new event-decoding logic | Not added — acknowledged not blocking. |
| 7 | Multi-line JSDoc vs CLAUDE.md | **Fixed.** Both functions now have one-line JSDoc. |

Bonus improvements I didn't ask for:
- **Explicit `chainId` on every `simulateContract` / `writeContract` / `waitForTransactionReceipt` call** — removes reliance on connector state, makes the cross-chain `addNominee` flow (current chain → mainnet) self-documenting.
- **Early guard** `if (chainId === undefined) throw new Error('Cannot determine chain ID')` replaces the previous `chainId as number` wishful typing.
- **The mock comment** in `index.spec.tsx` now explains *why* the mock exists (ESM transform limit), which is exactly the kind of WHY-comment CLAUDE.md asks for.

## Verdict

LGTM — clean, focused PR. Approving.

rajat2502 (approved):
LGTM. Clean, well-scoped migration. Two things to keep an eye on post-merge:

1. **Safe-via-WalletConnect path**: dropping the `provider.getCode(account)` Safe-detection in the old `sendTransaction` wrapper means `waitForTransactionReceipt` will hang on a Safe tx still awaiting co-signers. The wagmi `safe()` connector covers the Safe-Apps iframe case but not WalletConnect-from-Safe-wallet. Worth a manual smoke test of both flows.

2. **UX change in Create**: old code redirected to the staking-contracts list even when `InstanceCreated` was absent; new code throws to `ErrorAlert`. Functionally cleaner — just confirm with product this is the desired post-migration behavior.

Nit (non-blocking): passing `eventName: 'InstanceCreated'` to `decodeEventLog` would let viem type `decoded.args` and drop the `as unknown as { instance: Address }` cast.

PR #397, opened 2026-05-19
chore(deps): migrate build's web3.ts code to viem

atepem:
PR #3 of the web3.js -> viem modernization. Build's contract helpers (getDispenserContract, getTreasuryContract, getTokenomicsContract, getAgentContract, getComponentContract) and getBlockTimestamp utility were the last web3.js consumers; they're dropped in favor of direct @wagmi/core calls.

apps/build/common-util/functions/index.ts
- Removed: getWeb3Details, getContract (internal), getDispenserContract, getTreasuryContract, getTokenomicsContract, getAgentContract, getComponentContract, getBlockTimestamp.
- Kept: getProvider, getChainId, getChainIdOrDefaultToMainnet, getIsValidChainId, sendTransaction (all wrap libs/util-functions and are independent of web3.js).
- Drops the web3 import entirely.

apps/build/components/DevIncentives/requests.ts
- Sole consumer of the deleted helpers. Rewritten end-to-end:
  - 10 reads (.methods.foo().call()) -> readContract(wagmiConfig, ...)
  - 2 writes (.methods.foo().send()) -> simulateContract + writeContract
    + waitForTransactionReceipt
  - getBlockTimestamp (web3.eth.getBlock) -> getBlock from @wagmi/core
- Per-contract `*Params(chainId)` helpers consolidate address/abi/chainId threading.
- bigint output handling: epochCounter / epochLen / inflationPerSecond now properly typed as bigint and coerced where needed (parseInt(str) -> Number(bigint)). Caller (getPendingIncentives) already used BigInt arithmetic, so calculations are unchanged.

apps/build/components/DevIncentives/ClaimIncentives.tsx
- `paused()` returns bool in Solidity. web3.js 1.x returned native boolean but typed it as `any` from `.call()`, so the local state `useState('0')` silently accepted boolean assignment. Viem's strict return typing catches the mismatch. Minimal fix preserves the existing (buggy) `pauseValue === '0'` comparison logic by coercing with String(value) at the assignment site; the latent UI logic bug is out of scope for this migration.

Eliminates the web3.js import from build/. Remaining: bond (2), contribute (2), govern (1), marketplace (1).

Verification on this branch
- yarn nx build build: green
- yarn supply-chain: green (1 allowlisted bigint-buffer, 0 unlisted)
- build app has no spec files, so no Jest mocks to add

## Types of changes

## Supply-chain checklist

- [x] `yarn supply-chain` passes locally.

### DIFF
--- a/apps/build/common-util/functions/index.ts
+++ b/apps/build/common-util/functions/index.ts
@@ -1,4 +1,3 @@
-import Web3 from 'web3';
 import { RPC_URLS } from 'libs/util-constants/src';
 import {
   getProvider as getProviderFn,
@@ -7,14 +6,8 @@ import {
   getIsValidChainId as getIsValidChainIdFn,
   sendTransaction as sendTransactionFn,
 } from 'libs/util-functions/src';
+
 import { SUPPORTED_CHAINS } from 'components/Login';
-import {
-  DISPENSER,
-  TREASURY,
-  TOKENOMICS,
-  AGENT_REGISTRY,
-  COMPONENT_REGISTRY,
-} from 'libs/util-contracts/src/lib/abiAndAddresses';
 
 type ChainId = number | string;
 
@@ -23,10 +16,8 @@ export const getProvider = () => getProviderFn(SUPPORTED_CHAINS, RPC_URLS);
 export const getIsValidChainId = (chainId: ChainId) =>
   getIsValidChainIdFn(SUPPORTED_CHAINS, chainId);
 
-export const getChainIdOrDefaultToMainnet = (chainId: ChainId) => {
-  const x = getChainIdOrDefaultToMainnetFn(SUPPORTED_CHAINS, chainId);
-  return x;
-};
+export const getChainIdOrDefaultToMainnet = (chainId: ChainId) =>
+  getChainIdOrDefaultToMainnetFn(SUPPORTED_CHAINS, chainId);
 
 export const getChainId = (chainId = null) => getChainIdFn(SUPPORTED_CHAINS, chainId);
 
@@ -40,79 +31,3 @@ export const sendTransaction = (
     supportedChains: SUPPORTED_CHAINS,
     rpcUrls: RPC_URLS,
   });
-
-/**
- * returns the web3 details
- */
-const getWeb3Details = () => {
-  const chainId = getChainId();
-  const web3 = new Web3(getProvider());
-  return { web3, chainId };
-};
-
-/**
- * returns the contract instance
- * @param {Array} abi - abi of the contract
- * @param {String} contractAddress - address of the contract
- */
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
-const getContract = (abi: any, contractAddress: string) => {
-  const { web3 } = getWeb3Details();
-  const contract = new web3.eth.Contract(abi, contractAddress);
-  return contract;
-};
-
-export const getDispenserContract = () => {
-  const { chainId } = getWeb3Details();
-  const contract = getContract(
-    DISPENSER.abi,
-    DISPENSER.addresses[chainId as unknown as keyof typeof DISPENSER.addresses],
-  );
-  return contract;
-};
-
-export const getTreasuryContract = () => {
-  const { chainId } = getWeb3Details();
-  const contract = getContract(
-    TREASURY.abi,
-    TREASURY.addresses[chainId as unknown as keyof typeof TREASURY.addresses],
-  );
-  return contract;
-};
-
-export const getTokenomicsContract = () => {
-  const { chainId } = getWeb3Details();
-  const contract = getContract(
-    TOKENOMICS.abi,
-    TOKENOMICS.addresses[chainId as unknown as keyof typeof TOKENOMICS.addresses],
-  );
-  return contract;
-};
-
-export const getAgentContract = () => {
-  const { chainId } = getWeb3Details();
-  const contract = getContract(
-    AGENT_REGISTRY.abi,
-    AGENT_REGISTRY.addresses[chainId as unknown as keyof typeof AGENT_REGISTRY.addresses],
-  );
-  return contract;
-};
-
-export const getComponentContract = () => {
-  const { chainId } = getWeb3Details();
-  const contract = getContract(
-    COMPONENT_REGISTRY.abi,
-    COMPONENT_REGISTRY.addresses[chainId as unknown as keyof typeof COMPONENT_REGISTRY.addresses],
-  );
-  return contract;
-};
-
-/**
- * TODO: move to autonolas-library and figure out a better way
- * to fetch timestamp
- */
-export const getBlockTimestamp = async (block = 'latest') => {
-  const { web3 } = getWeb3Details();
-  const temp = await web3.eth.getBlock(block);
-  return Number(temp.timestamp);
-};
--- a/apps/build/components/DevIncentives/ClaimIncentives.tsx
+++ b/apps/build/components/DevIncentives/ClaimIncentives.tsx
@@ -18,8 +18,10 @@ export const ClaimIncentives = () => {
   useEffect(() => {
     const getData = async () => {
       try {
+        // Treasury `paused()` returns a uint8 (0/1/2 — three pause states);
+        // `pauseValue` is compared as a string downstream (`=== '0' / '1' / '2'`).
         const value = await getPausedValueRequest();
-        setPausedValue(value);
+        setPausedValue(String(value));
       } catch (error: unknown) {
         notifySpecificError(error as Error);
         console.error(error);
--- a/apps/build/components/DevIncentives/IncentivesForThisEpoch.tsx
+++ b/apps/build/components/DevIncentives/IncentivesForThisEpoch.tsx
@@ -115,8 +115,8 @@ export const IncentivesForThisEpoch = () => {
             setRewardAndTopUp([
               {
                 key: '1',
-                reward: round(Number(parseToEth(response.reward)), 6),
-                topUp: round(Number(parseToEth(response.topUp)), 6),
+                reward: round(Number(parseToEth(response.reward.toString())), 6),
+                topUp: round(Number(parseToEth(response.topUp.toString())), 6),
               },
             ]);
           } catch (error: unknown) {
--- a/apps/build/components/DevIncentives/requests.ts
+++ b/apps/build/components/DevIncentives/requests.ts
@@ -1,19 +1,69 @@
-import { notifyError } from 'libs/util-functions/src';
-import { TOKENOMICS_UNIT_TYPES } from 'libs/util-constants/src';
-import { rewardsFormatter } from 'libs/common-contract-functions/src/lib/utils';
 import {
-  getAgentContract,
-  getBlockTimestamp,
-  getComponentContract,
-  getDispenserContract,
-  getTokenomicsContract,
-  getTreasuryContract,
-  sendTransaction,
-} from 'common-util/functions';
+  getBlock,
+  readContract,
+  simulateContract,
+  waitForTransactionReceipt,
+  writeContract,
+} from '@wagmi/core';
+import { Abi, Address } from 'viem';
 
-/**
- * fetches the owners of the units
- */
+import { rewardsFormatter } from 'libs/common-contract-functions/src/lib/utils';
+import { TOKENOMICS_UNIT_TYPES } from 'libs/util-constants/src';
+import {
+  AGENT_REGISTRY,
+  COMPONENT_REGISTRY,
+  DISPENSER,
+  TOKENOMICS,
+  TREASURY,
+} from 'libs/util-contracts/src/lib/abiAndAddresses';
+import { notifyError } from 'libs/util-functions/src';
+
+import { wagmiConfig } from 'components/Login/config';
+
+import { getChainId } from 'common-util/functions';
+
+const requireChainId = (): number => {
+  const chainId = getChainId();
+  if (chainId === undefined) throw new Error('Cannot determine chain ID');
+  return chainId;
+};
+
+const tokenomicsParams = (chainId: number) => ({
+  address: TOKENOMICS.addresses[chainId as keyof typeof TOKENOMICS.addresses] as Address,
+  abi: TOKENOMICS.abi,
+  chainId,
+});
+
+const treasuryParams = (chainId: number) => ({
+  address: TREASURY.addresses[chainId as keyof typeof TREASURY.addresses] as Address,
+  abi: TREASURY.abi,
+  chainId,
+});
+
+const dispenserParams = (chainId: number) => ({
+  address: DISPENSER.addresses[chainId as keyof typeof DISPENSER.addresses] as Address,
+  abi: DISPENSER.abi,
+  chainId,
+});
+
+// agent / component registry ABIs are not declared `as const` in the source `.js`
+// files, so viem can't infer return types — `as Abi` is required and the
+// `ownerOf` return must be cast manually.
+const agentParams = (chainId: number) => ({
+  address: AGENT_REGISTRY.addresses[chainId as keyof typeof AGENT_REGISTRY.addresses] as Address,
+  abi: AGENT_REGISTRY.abi as Abi,
+  chainId,
+});
+
+const componentParams = (chainId: number) => ({
+  address: COMPONENT_REGISTRY.addresses[
+    chainId as keyof typeof COMPONENT_REGISTRY.addresses
+  ] as Address,
+  abi: COMPONENT_REGISTRY.abi as Abi,
+  chainId,
+});
+
+/** Fetches the owners of the units. */
 export const getOwnersForUnits = async ({
   unitIds,
   unitTypes,
@@ -21,25 +71,23 @@ export const getOwnersForUnits = async ({
   unitIds: string[];
   unitTypes: string[];
 }) => {
-  const ownersList = [];
+  const chainId = requireChainId();
 
-  const agentContract = getAgentContract();
-  const componentContract = getComponentContract();
-
-  for (let i = 0; i < unitIds.length; i += 1) {
+  const ownersList = unitIds.map((unitId, i) => {
     // 1 = agent, 0 = component
-    if (unitTypes[i] === TOKENOMICS_UNIT_TYPES.AGENT) {
-      const result = agentContract.methods.ownerOf(unitIds[i]).call();
-      ownersList.push(result);
-    } else {
-      const result = componentContract.methods.ownerOf(unitIds[i]).call();
-      ownersList.push(result);
-    }
-  }
+    const params =
+      unitTypes[i] === TOKENOMICS_UNIT_TYPES.AGENT
+        ? agentParams(chainId)
+        : componentParams(chainId);
 
-  const list = await Promise.all(ownersList);
-  const results = await Promise.all(list.map((e) => e));
-  return results;
+    return readContract(wagmiConfig, {
+      ...params,
+      functionName: 'ownerOf',
+      args: [BigInt(unitId)],
+    }) as Promise<Address>;
+  });
+
+  return Promise.all(ownersList);
 };
 
 export const getOwnerIncentivesRequest = async ({
@@ -51,11 +99,13 @@ export const getOwnerIncentivesRequest = async ({
   unitTypes: string[];
   unitIds: string[];
 }) => {
-  const contract = getTokenomicsContract();
-  const response: { reward: string; topUp: string } = await contract.methods
-    .getOwnerIncentives(address, unitTypes, unitIds)
-    .call();
-  return response;
+  const chainId = requireChainId();
+  const [reward, topUp] = await readContract(wagmiConfig, {
+    ...tokenomicsParams(chainId),
+    functionName: 'getOwnerIncentives',
+    args: [address as Address, unitTypes.map(BigInt), unitIds.map(BigInt)],
+  });
+  return { reward, topUp };
 };
 
 export const claimOwnerIncentivesRequest = async ({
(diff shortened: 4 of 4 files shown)


mohandast52:
Reviewed in depth. **Do not merge as-is** — there are multiple runtime-breaking bugs that `yarn nx build build` cannot catch, because the build app has no spec files and TypeScript's strict types were laundered through `as Abi` + manual casts.

I verified the viem decoding behavior against the actual ABI by running decode/encode in Node:

```
mapUnitIncentives result: [ 1n, 22n, 333n, 4444n, 55555 ]
  isArray: true
  result.pendingRelativeReward: undefined

getOwnerIncentives: [ 100n, 200n ]
  isArray: true
  r2.reward: undefined

paused (uint8): 1, typeof: number
getUnitPoint.rewardUnitFraction typeof: number   // uint8
getUnitPoint.sumUnitTopUpsOLAS typeof: bigint     // uint96
100n * 80 THROWS: Cannot mix BigInt and other types
```

This rules in three runtime-breaking bugs (A–C) and a couple of misleading type casts (D, E).

---

### A. `getOwnerIncentives` is an array, not an object — `response.reward` / `response.topUp` are `undefined`

[apps/build/components/DevIncentives/requests.ts:100-104](apps/build/components/DevIncentives/requests.ts#L100-L104) casts the read as `{ reward: string; topUp: string }`. The Tokenomics ABI defines `getOwnerIncentives` with **two top-level outputs** (`reward: uint256`, `topUp: uint256`) — not a struct. Viem decodes multi-top-level outputs as a tuple/array, so the actual runtime shape is `[bigint, bigint]`. The downstream code in [IncentivesForThisEpoch.tsx:118-119](apps/build/components/DevIncentives/IncentivesForThisEpoch.tsx#L118-L119) reads `response.reward` and `response.topUp` — both undefined at runtime. The "Check for available rewards" feature is broken.

Fix:
```ts
const [reward, topUp] = (await readContract(wagmiConfig, {
  ...tokenomicsParams(chainId),
  functionName: 'getOwnerIncentives',
  args: [address, unitTypes, unitIds],
})) as readonly [bigint, bigint];
return { reward, topUp };
```

### B. `mapUnitIncentives` is an array, not an object — destructured fields are `undefined`

Same root cause. [requests.ts:231-237](apps/build/components/DevIncentives/requests.ts#L231-L237) casts as `{ pendingRelativeReward, pendingRelativeTopUp, lastEpoch }`, but `mapUnitIncentives` returns **5 top-level outputs** (uint96, uint96, uint96, uint96, uint32). Viem returns `[bigint, bigint, bigint, bigint, number]`. The destructure at [requests.ts:237](apps/build/components/DevIncentives/requests.ts#L237) produces all-undefined values, so `isCurrentEpochWithReward` short-circuits to `false` for everyone, and the entire `getPendingIncentives` function silently always returns zero rewards.

Fix:
```ts
const [, pendingRelativeReward, , pendingRelativeTopUp, lastEpoch] =
  (await readContract(wagmiConfig, { ...tokenomicsParams(chainId), functionName: 'mapUnitIncentives', args: [unitType, unitId] })) as readonly [bigint, bigint, bigint, bigint, number];
```

Compare to govern's [hooks/useNomineesWeights.ts](apps/govern/hooks/useNomineesWeights.ts) and to govern's existing ethers-based [libs/common-contract-functions/src/lib/rewards.ts:46-54](libs/common-contract-functions/src/lib/rewards.ts#L46-L54) — that file destructures named fields because **ethers v6 Results expose both indexed and named access; viem does not**. The migration from ethers/web3.js to viem needs explicit tuple destructuring.

### C. `bigint * number` in `getPendingIncentives` arithmetic throws at runtime

Even if A/B are fixed, the arithmetic on [requests.ts:257, 275, 279](apps/build/components/DevIncentives/requests.ts#L257-L279) mixes bigint with number:

- `pendingRelativeReward * unitInfo.rewardUnitFraction` — `bigint × number` (uint96 × uint8) → `TypeError: Cannot mix BigInt and other types, use explicit conversions`
- `totalIncentives * unitInfo.topUpUnitFraction` — same
- `unitInfo.sumUnitTopUpsOLAS * BIG_INT_100` — OK (bigint × bigint)

`getUnitPoint` returns a struct, so the object-access works, but viem returns uint8 fields as `number` and uint96 fields as `bigint`. The pre-migration web3.js code stringified everything, so the explicit `BigInt(unitInfo.rewardUnitFraction)` wrappers in the original were load-bearing — the PR removed them. Restore them:

```ts
const pendingReward =
  (pendingRelativeReward * BigInt(unitInfo.rewardUnitFraction)) / BIG_INT_100;
...
pendingTopUp =
  (totalIncentives * BigInt(unitInfo.topUpUnitFraction)) /
  (unitInfo.sumUnitTopUpsOLAS * BIG_INT_100);
```

### D. `paused()` is `uint8`, not `bool` — the PR description has this wrong

[requests.ts:194-200](apps/build/components/DevIncentives/requests.ts#L194-L200) casts as `boolean`. The Treasury ABI declares `paused` with output `internalType: 'uint8'`, viem decodes it as `number`. The UI logic in [ClaimIncentives.tsx:58-89](apps/build/components/DevIncentives/ClaimIncentives.tsx#L58-L89) checks `pauseValue === '0'`, `'1'`, `'2'` — **three states**, not two. That is intentional Treasury behavior (0 = unpaused, 1 = paused for claiming only, 2 = fully paused). It is not a "latent UI logic bug" as the PR description claims.

`String(number)` at the assignment site happens to round-trip the three states ('0', '1', '2'), so the runtime is unchanged — but please fix the cast to `as number` and update the PR description so the "latent bug" framing doesn't propagate into a follow-up cleanup PR that would actually break the three-state UI.

### E. `mapEpochTokenomics.endTime` typed as `bigint`, actually `number`

[requests.ts:155](apps/build/components/DevIncentives/requests.ts#L155) — `endTime` is uint32 → `number`. The runtime works because `Number(number) === number`, but the type is wrong.

---

### Why `yarn nx build build` and `yarn supply-chain` are green

The `as Abi` cast on every contract config erases viem's strict return-type inference, then the manual `as { … }` casts on every `readContract` result satisfy TypeScript. Type-checking has no signal here; execution is the only way to catch A/B/C.

If you instead let viem infer types from the ABI:

```ts
const tokenomicsParams = (chainId: MaybeChainId) => ({
  address: TOKENOMICS.addresses[chainId as keyof typeof TOKENOMICS.addresses] as Address,
  abi: TOKENOMICS.abi,                  // not `as Abi`
  chainId: chainId as number,
} as const);
```

…then viem would have rejected the object-destructure on `mapUnitIncentives` at compile time, the `as boolean` on `paused`, etc. The ABIs in [libs/util-contracts/src/lib/abiAndAddresses](libs/util-contracts/src/lib/abiAndAddresses/) are already declared as `as const`, so this should work — worth doing for this whole DevIncentives module.

---

### Verdict

**Do not approve.** A, B, C are user-visible regressions: "Check for available rewards" returns garbage, claim/checkpoint paths would still write but reward display lies, and pending-incentives throws on any unit with a non-zero balance. The build app has no tests, so "build green + supply-chain green" is not adequate verification — this needs a manual smoke against mainnet (or a fork) of:

- DevIncentives → "Check for available rewards" → enter a known-owner address + unit IDs → confirm `reward`/`topUp` columns populate correctly.
- DevIncentives → "Claim Rewards" with `paused` = 1 → confirm form renders, transaction simulates.
- DevIncentives → "Check Pending Incentives" → confirm `pendingReward`/`pendingTopUp` calculate without throwing.

Bonus suggestion: drop `as Abi` and let viem infer return types — the same class of bug will keep happening across the other web3.js → viem PRs (#395 you already merged-targeting `precious/viem-migration`, plus bond/contribute/marketplace still to come) unless the typing is tightened.

Migration shape (per-contract `*Params` helpers, the `simulateContract` → `writeContract` → `waitForTransactionReceipt` pattern, dropping `sendTransaction` for writes) is otherwise good — the bugs are concentrated in the read-side type contract, not the structure of the rewrite.

Tanya-atatakai (commented):
# PR Review: `chore(deps): migrate build's web3.ts code to viem`

## Overview

**Declared scope:** Replace `apps/build/common-util/functions/index.ts`'s web3.js Contract helpers with direct `@wagmi/core` calls; rewrite `apps/build/components/DevIncentives/requests.ts` accordingly; coerce a `boolean` state in `ClaimIncentives.tsx`.

**Actual scope** (per `git diff --stat`): the above, **plus** `.supply-chain/audit-allowlist.json`, `.supply-chain/install-hooks.allowlist`, `nx.json`, `package.json`, and 642 lines of `yarn.lock` churn. Same scope-creep pattern as #395 (minus the proposal doc).

The migration itself is well-executed and noticeably cleaner than the web3.js original.

---

## ✅ What's good

- **Per-contract `*Params(chainId)` helpers** are a nice DRY refactor — address/abi/chainId threaded through a single object. Reads idiomatically.
- **`getOwnersForUnits` is genuinely cleaner.** The old loop did `await Promise.all(list.map((e) => e))` after a first `Promise.all` — redundant and confusing. The new `.map` + `Promise.all` is straightforward.
- **bigint arithmetic propagated correctly** through `getPendingIncentives`:
  - `pendingRelativeReward * unitInfo.rewardUnitFraction` (was wrapped in `BigInt(...)` defensively before; now native).
  - `inflationPerSecond * BigInt(epochLength)` explicitly coerces the `number` to bigint — good.
  - `totalIncentives * unitInfo.topUpUnitFraction / (unitInfo.sumUnitTopUpsOLAS * BIG_INT_100)` — pure bigint, no string coercion needed.
- **`epochCounter` / `epochLen` / `inflationPerSecond` properly typed `bigint`** with `Number(...)` coercion at the boundary. Old `parseInt(string, 10)` is gone.
- **Idiomatic write pattern**: `simulateContract` → `writeContract` → `waitForTransactionReceipt`. Matches the launch PR.
- **No orphaned callers**: `getOwnerIncentivesRequest` and `getOwnersForUnits` are consumed in `IncentivesForThisEpoch.tsx` — both signatures preserved.

---

## ⚠️ Issues

### 1. Scope creep (same as #395)

The diff also includes:
- `nx.json` — re-introduces `build-storybook` and `e2e` task targets and `cacheableOperations: ["build-storybook"]`. Commit `de4eeacd` on main just *removed* Storybook tooling — this looks like a stale-rebase revert. **Same suspicious pattern as #395.**
- `.supply-chain/audit-allowlist.json` (bigint-buffer entry), `install-hooks.allowlist` (bigint-buffer + cypress).
- `package.json` (removes `@balancer-labs/sdk`, `@particle-network/auth`, `@web3inbox/*`).
- 642 lines of `yarn.lock` churn (vs ~40 lines that are unique to this branch over the launch branch — the rest is shared scope creep).

`web3-to-viem-launch` and `web3-to-viem-build` are not stacked (`git merge-base --is-ancestor` → NO) — each independently carries these shared changes. If only one of #395/#397 lands, the other still has the same scope. **Rebase one onto the other, or split out the shared changes into their own PR.**

### 2. **Type cast on `getOwnerIncentivesRequest` is dishonest** (works at runtime, but lies to TS)

```ts
const response = (await readContract(wagmiConfig, {...})) as { reward: string; topUp: string };
```

Solidity returns `(uint256 reward, uint256 topUp)` — viem returns these as **`bigint`**, not `string`. The cast claims `string`. It happens to work downstream because the consumer is:

```ts
// IncentivesForThisEpoch.tsx:112
reward: round(Number(parseToEth(response.reward)), 6),
```

…and `parseToEth(amount: string) => formatEther(\`${amount}\`)` (`libs/util-functions/src/lib/ethers.ts:22`) template-literal-coerces the bigint to a decimal string before `formatEther` parses it. Works by accident.

**Fix:** type as `bigint` and update `parseToEth`'s signature, or coerce explicitly:

```ts
}) as { reward: bigint; topUp: bigint };
// ...
reward: round(Number(parseToEth(response.reward.toString())), 6),
```

This same shape applies to `unitIncentives` (`pendingRelativeReward: bigint`) — there the typing IS correct, so the inconsistency makes the `getOwnerIncentives` cast stand out more.

### 3. **`MaybeChainId = number | undefined` then `chainId as number` is wishful typing**

```ts
const tokenomicsParams = (chainId: MaybeChainId) => ({
  address: TOKENOMICS.addresses[chainId as unknown as keyof typeof TOKENOMICS.addresses] as Address,
  abi: TOKENOMICS.abi as Abi,
  chainId: chainId as number,  // ← if undefined at runtime, fails opaquely
});
```

If `getChainId()` ever returns `undefined` (disconnected, unsupported chain), `TOKENOMICS.addresses[undefined]` is `undefined`, viem sees `address: undefined as Address`, and the error surfaces deep in the wagmi stack. The build app's `SUPPORTED_CHAINS = [mainnet]` makes this near-unreachable in practice, but worth either:
- Throwing early at the top of each request: `if (!chainId) throw new Error('No chain selected');`
- Or making `chainId` non-optional in the helpers and validating at the function entry.

### 4. **`paused === '0'` comparison is a pre-existing bug — please track it**

PR description acknowledges this. For the record: `pauseValue` state holds `'true'` or `'false'` after `String(value)`, but the JSX compares `pauseValue === '0'` / `'1'` / `'2'`. **None of those ever match** — meaning the "Claim Rewards" UI was always hidden. With web3.js it was the same (boolean coerced to whatever `===` did with `any`). The `String(value)` fix preserves the bug as-is; OK as scope, but please file a follow-up — this is a real UX issue, not just a typo.

### 5. **Comment style vs CLAUDE.md** (minor nit)

CLAUDE.md: *"Never write multi-paragraph docstrings or multi-line comment blocks — one short line max."* The retained `/** fetches the owners of the units */` is fine; the `/** the below calculation is done to get the reward and topup based on current epoch length */` 3-line block is technically a violation. Leftover from the pre-existing code, not introduced here.

### 6. No tests

Per the PR body, "build app has no spec files." That's accurate, but `getPendingIncentives` is a substantial bigint-arithmetic flow worth at least one unit test (mock `@wagmi/core`, feed synthetic responses, assert the output formatting). Not blocking, but the build app's complete lack of test coverage is a separate problem that this PR neither helps nor harms.

---

## 🔒 Security / supply chain

- `yarn supply-chain` checkbox unchecked in template, but the PR touches `.supply-chain/` and `package.json`. Confirm a local run passed.
- Shared with #395: `bigint-buffer` audit allowlist entry is well-justified (no upstream patch, 3-month review).

---

## Verdict

The build migration itself is good — arguably cleaner than launch's, because the `*Params` helper pattern is a nice generalization. **Approve once:**
1. The scope-creep is split out (or confirmed intentional and the `nx.json` "Storybook revert" is verified against main).
2. The `getOwnerIncentivesRequest` return-type cast is corrected (`string` → `bigint` + `.toString()` at the call site).
3. A follow-up issue is filed for the latent `pauseValue === '0'` UI bug.

Items 1 and 2 are the only blockers; 3 is housekeeping.

Tanya-atatakai:
Re-review check: this PR doesn't appear to have been updated since my prior review — same two commits (`25cd5759`, `0c6224e3`), same 8-file diff (799+/202- including the scope-creep files: `.supply-chain/*`, `nx.json`, `package.json`, `yarn.lock`).

Could you push the addressed changes? Once they're up I'll re-review against the prior comments (esp. the `getOwnerIncentivesRequest` cast as `string` vs viem's actual `bigint`, the `MaybeChainId` early guard, and confirming the `nx.json` Storybook config isn't reintroduced).

atepem:
| Reviewer concern | Status |
|---|---|
| Scope creep — `nx.json` / `package.json` / `yarn.lock` / supply-chain deltas | ✅ Auto-resolved by rebasing onto current umbrella (was rebase staleness, not my changes) |
| A. `getOwnerIncentives` is array not object → `response.reward` undefined | ✅ Tuple destructure `[reward, topUp]` then return `{ reward, topUp }` |
| B. `mapUnitIncentives` is 5-tuple, not object → all destructured fields undefined | ✅ Tuple destructure `[, pendingRelativeReward, , pendingRelativeTopUp, lastEpoch]` |
| C. `bigint × number` throws in `getPendingIncentives` arithmetic | ✅ Restored `BigInt(unitInfo.rewardUnitFraction)` and `BigInt(unitInfo.topUpUnitFraction)` wrappers |
| D. `paused()` is `uint8` (3-state), not `bool` | ✅ Return type changed to `number`; `ClaimIncentives` comment updated to reflect intentional three-state UI |
| E. `mapEpochTokenomics.endTime` typed as `bigint`, actually `number` | ✅ Removed manual type cast; viem now infers correctly from the `const` ABI |
| Copilot: `ownerOf(unitId)` needs `BigInt(unitId)` | ✅ Fixed |
| Copilot: `getOwnerIncentives` arrays need `bigint[]` | ✅ `unitTypes.map(BigInt)`, `unitIds.map(BigInt)` |
| Copilot: `mapUnitIncentives` `unitType`/`unitId` need `bigint` | ✅ `BigInt(unitType)`, `BigInt(unitId)` |
| Copilot: `getUnitPoint` `unitType` needs `bigint` | ✅ Fixed |
| `as Abi` strips type inference | ✅ Dropped for `TOKENOMICS`, `TREASURY`, `DISPENSER` (which are `as const`); kept for `AGENT_REGISTRY` / `COMPONENT_REGISTRY` (plain `.js` without `as const`, conversion is out of scope) |
| `getOwnerIncentivesRequest as { reward: string; topUp: string }` lies | ✅ Returns proper `{ reward: bigint; topUp: bigint }`. Caller in `IncentivesForThisEpoch.tsx` updated with `.toString()` for `parseToEth` |
| `MaybeChainId` silent `undefined` | ✅ `requireChainId()` helper throws early if undefined |
| PR description's "latent bug" framing for `paused` | ✅ Corrected in commit message (was wrong; three-state UI is intentional) |
| No tests for `getPendingIncentives` | ❌ Build app has no spec files (pre-existing); flagging in PR body for follow-up |
| `pauseValue === '0'` UI follow-up bug | ❌ Out of scope — wait, this is actually NOT a bug now that we know `paused()` is `uint8`. The three-state comparison is correct |

The key insight from the reviewer was right: dropping `as Abi` for the `as const` ABIs catches all these bugs at compile time. The agent/component registry JS files still need conversion to TS with `as const` for full coverage, but that's a follow-up.

Tanya-atatakai (approved):
# Re-review — all prior feedback addressed ✅

Force-push between v1 and now collapsed two commits into one (`a6475703`) and stripped all the scope-creep. Diff is now: 4 files, all in `apps/build/`. Stats 177/172 vs prior 799/202.

## Each prior comment, verified

| # | Prior feedback | Status |
|---|---|---|
| 1 | Scope creep (`.supply-chain/*`, `nx.json`, `package.json`, 642 lines of `yarn.lock`) | **Removed.** Diff is now exactly the migration. |
| 2 | `getOwnerIncentivesRequest` cast as `{ reward: string; topUp: string }` but viem returns `bigint` | **Fixed properly.** Now destructures the tuple `[reward, topUp]` from the inferred ABI return and the caller in `IncentivesForThisEpoch.tsx` adds `.toString()` before `parseToEth`. The fix is at both ends — type is honest and the runtime path is explicit. |
| 3 | `MaybeChainId = number \| undefined` + `chainId as number` wishful typing | **Fixed.** Replaced with a `requireChainId()` helper that throws explicitly if `getChainId()` returns undefined. Used uniformly across every request function. Much cleaner than the per-call cast. |
| 4 | `pauseValue === '0'` latent UI bug, deferred | **Actually fixed.** Recognised that Solidity `paused()` is a **uint8 with 3 states** (0 = unpaused, 1 = paused for claiming only, 2 = fully paused), not a `bool`. `getPausedValueRequest` now returns `Number(response)`; `ClaimIncentives.tsx` does `setPausedValue(String(value))` and the downstream `=== '0' / '1' / '2'` comparisons will now actually work. This is the bonus the original PR explicitly deferred — nice to see it fixed in the same PR. |
| 5 | Multi-line block comment in `getPendingIncentives` | **Replaced with targeted WHY-comments** that are exactly the kind CLAUDE.md asks for: |

## Bonus improvements I didn't ask for

- **`*Params` helpers no longer cast `abi as Abi`** for `tokenomics`/`treasury`/`dispenser` — the inferred return types from those typed ABIs (`.ts` files) flow through to call sites, removing several redundant casts. The `as Abi` is kept only where actually needed (agent/component registries are `.js` files, not `as const`) with an inline comment explaining why.
- **`unitIds.map(BigInt)`, `BigInt(unitId)`, `BigInt(epochNum)`**, etc. — explicit `BigInt()` wrappers on every uint256 arg rather than relying on viem's implicit string coercion. More correct and self-documenting.
- **`mapUnitIncentives` tuple destructure** with a precise comment:
  > `mapUnitIncentives returns 5 top-level outputs (uint96, uint96, uint96, uint96, uint32), so viem decodes it as a tuple — not an object with named fields.`
  
  The destructure `const [, pendingRelativeReward, , pendingRelativeTopUp, lastEpoch] = unitIncentives;` correctly skips the two unused fields by position. This is a viem footgun ([their docs flag it](https://viem.sh/docs/contract/readContract.html#return-value)) and the comment will save the next reader 20 minutes.
- **`getUnitPoint` mixed-type arithmetic comment**:
  > `getUnitPoint returns a struct with mixed uint8 (number) and uint96 (bigint) fields. The uint8 fields must be wrapped in BigInt() before multiplying with uint96 bigints — mixing throws "Cannot mix BigInt and other types".`
  
  Exactly the WHY-comment style CLAUDE.md asks for.
- **`endTime` typing comment**:
  > `` `endTime` is a uint32 — viem returns it as `number` from the inferred ABI shape. ``
- **JSDoc trimmed** to one-liner per CLAUDE.md style guide.

## Minor things I noticed

- `simulateContract` includes `chainId` via the spread but `writeContract` / `waitForTransactionReceipt` do not. This is fine for build (single-chain `[mainnet]`) and follows the wagmi pattern of letting the `request` object from simulate carry the chain context. Not a blocker — just noting the difference from launch's PR which explicitly threads `chainId` because it crosses chains.
- `address: address as Address` cast remains in `getOwnerIncentives` — same loose-typing pattern as elsewhere in the workspace. Fine.

## Verdict

LGTM. Approving — the migration is now equivalent to launch's quality plus a real bonus bug fix.

rajat2502 (approved):

mohandast52 (approved):
Re-reviewed against the new head (a647570306). All five issues from my previous comment ([#issuecomment-4488285110](https://github.com/valory-xyz/autonolas-frontend-mono/pull/397#issuecomment-4488285110)) are addressed:

### Fixed
- **A — `getOwnerIncentives` tuple destructuring** ✅ `const [reward, topUp] = await readContract(...)` — verified against viem decode, returns `[100n, 200n]` for `[uint256, uint256]`. Consumer in [IncentivesForThisEpoch.tsx:118-119](apps/build/components/DevIncentives/IncentivesForThisEpoch.tsx#L118-L119) updated to call `.toString()` on the bigint values.
- **B — `mapUnitIncentives` tuple destructuring** ✅ `const [, pendingRelativeReward, , pendingRelativeTopUp, lastEpoch] = unitIncentives` — index pattern is correct for the 5-element tuple `[reward, pendingRelativeReward, topUp, pendingRelativeTopUp, lastEpoch]`. Verified against actual decode.
- **C — `bigint × number` arithmetic** ✅ Both `BigInt(unitInfo.rewardUnitFraction)` and `BigInt(unitInfo.topUpUnitFraction)` wrappers restored at [requests.ts:267](apps/build/components/DevIncentives/requests.ts#L267) and [requests.ts:285](apps/build/components/DevIncentives/requests.ts#L285). No more `Cannot mix BigInt and other types`.
- **D — `paused()` cast** ✅ Now `return Number(response)` with an explanatory comment about the three-state uint8 (0 = unpaused, 1 = paused for claiming only, 2 = fully paused). Downstream `String(value)` round-trip in ClaimIncentives.tsx is preserved, with a comment explaining the contract.
- **E — `mapEpochTokenomics.endTime` typing** ✅ `as Abi` dropped for Tokenomics, so viem now infers `endTime` as `number`. The inline `Number(epTokenomics.endTime)` is left in place with a comment.

### Bonus improvements
- **`as Abi` dropped** for the three contracts whose ABIs are declared `as const` in `.ts` files (Tokenomics, Treasury, Dispenser) — viem now infers return types. Agent/Component registry ABIs keep `as Abi` because they're in `.js` files (no `as const`), and the [requests.ts:49-51](apps/build/components/DevIncentives/requests.ts#L49-L51) comment captures *why*. Solid documentation of an otherwise non-obvious choice.
- **`requireChainId()` helper** ([requests.ts:25-29](apps/build/components/DevIncentives/requests.ts#L25-L29)) throws explicitly instead of silently fanning `undefined` through `chainId as number` casts. Strictly better than what was there.
- **uint256 args wrapped in `BigInt()`** throughout (`BigInt(unitId)`, `BigInt(unitType)`, `unitTypes.map(BigInt)`, `unitIds.map(BigInt)`, `BigInt(epochNum)`). This matches what we should have done from the start; sets a good convention for the remaining bond/contribute/marketplace PRs.
- **Comments where they matter**: every non-obvious decoding/coercion choice now has a WHY comment that captures the contract behavior. Future maintainers won't re-introduce the same bugs.

### Verified
- viem decode test (locally, against actual ABI shapes) — tuple destructuring picks the correct indices.
- Lint clean on the build app.
- PR description claims build/supply-chain green.
- No remaining `web3.js` import in `apps/build/`.

Approving. Thanks for the thorough fix — the explanatory comments make this a reference for future viem migrations.

Three down, two to go (bond, contribute, marketplace).

PR #399, opened 2026-05-19
chore(deps): migrate govern's web3.ts code to viem

atepem:
PR #4 of the web3.js -> viem modernization. Govern was the write-heaviest app to migrate (11 writes, 2 reads, all routed through libs/util-functions/sendTransaction with the dual-stack methodFn/Contract type chokepoint). All replaced with wagmi-native simulateContract + writeContract + waitForTransactionReceipt.

apps/govern/common-util/functions/web3.ts
- Deleted. The 6 mainnet-only contract getters (getVoteWeightingContract, getOlasContract, getVeOlasContract, getTokenomicsContract, getTreasuryContract, getGovernorContract) had exactly one consumer (requests.ts), and that consumer no longer needs them.

apps/govern/common-util/functions/index.ts
- Drops the `export * from './web3'` barrel re-export so the deletion doesn't break the barrel. No other consumer of the barrel uses any web3.ts export (verified by grep).

apps/govern/common-util/functions/requests.ts
- 2 reads (getAllRemovedNominees, allowance) -> readContract(wagmiConfig).
- 11 writes (voteForNomineeWeightsBatch, approve, createLock, increaseAmount, increaseUnlockTime, withdraw, checkpoint, depositServiceDonationsETH, castVote, revokeRemovedNomineeVotingPower) -> simulateContract + writeContract + waitForTransactionReceipt.
- Per-contract params constants (voteWeightingParams, olasParams, veOlasParams, tokenomicsParams, treasuryParams, governorParams) consolidate the { address, abi, chainId } triple. All target mainnet, matching the original behavior.
- depositServiceDonationsETH: BigInt(totalAmount) for the payable `value` field (web3.js auto-converted strings; viem requires bigint).
- Caller return-shape compatibility verified: only one caller (RevokePower.tsx) reads result.transactionHash; the field name is preserved on the new TransactionReceipt return. Other callers either await or .then(() => ...) without reading the return.
- Drops imports of getEstimatedGasLimit and sendTransaction from libs/util-functions; those library functions remain for the apps that still depend on them (bond, contribute, marketplace).

Eliminates the web3.js import from govern/. Remaining: bond (2), contribute (2), marketplace (1).

Verification on this branch
- yarn nx build govern: green
- yarn nx test govern: 10 suites, 57 tests pass
- npx eslint on changed files: clean

## Types of changes

## Supply-chain checklist


Tanya-atatakai (approved):
# PR Review: `chore(deps): migrate govern's web3.ts code to viem`

## Overview

The largest of the four web3.js→viem migrations — 11 writes + 2 reads, all routed through the previous `methodFn`/Contract bridge. Diff is 3 files (146/220), all in `apps/govern/common-util/functions/`. No scope creep. 👍

## ✅ Verified

- **`web3.ts` deletion is safe.** All 6 contract getters (`getVoteWeightingContract`, `getOlasContract`, `getVeOlasContract`, `getTokenomicsContract`, `getTreasuryContract`, `getGovernorContract`) had exactly one consumer (`requests.ts`); `git grep` against the PR branch confirms zero remaining callers. Barrel re-export in `index.ts` removed in the same commit.
- **Caller return-shape compatibility.** `RevokePower.tsx:132–140` reads `result?.transactionHash` — preserved because `revokePower` still returns the full `TransactionReceipt` (just from `waitForTransactionReceipt` instead of the old `sendTransaction` wrapper). Other callers (`ApproveOlasModal`, `Votes`, `ProposalsList`, `EditVotes`) either `await` the call or chain `.then(() => ...)` without reading the return — all safe.
- **Per-function return type preserved.** Half the writes return the full receipt (`voteForNomineeWeights`, `approveOlasByOwner`, `voteForProposal`, `revokePower`), half return just `receipt.transactionHash` (`createLockRequest`, `updateIncreaseAmount`, `updateIncreaseUnlockTime`, `withdrawVeolasRequest`, `checkpointRequest`, `depositServiceDonationRequest`). This **matches the prior behavior** of each function exactly — no caller will notice.
- **`depositServiceDonationsETH` payable `value`.** `BigInt(totalAmount)` is the right conversion: web3.js auto-coerced strings, viem requires bigint for the `value` field. Spot-checked.

## ✅ What's good

- **Top-level `*Params` constants** (vs build's per-call factory functions) — appropriate here because govern is mainnet-only and the `chainId` is a literal. Cleaner. The inline `const MAINNET = mainnet.id` + comment "All governance contracts live on mainnet." is a nice WHY-comment.
- **DRY win for pre-existing reads.** `checkNegativeSlope`, `checkLockExpired`, `checkServicesTerminatedOrNotDeployed` already used `readContract` directly with inline `address`/`abi`/`chainId` triples. The PR sweeps those into the shared `veOlasParams` / `MAINNET` constants in the same diff. Nice bundling of cleanup with the core migration.
- **`SUPPORTED_CHAINS` import dropped** from `requests.ts` — no longer needed since the wagmi-native calls don't need the old `sendTransaction` config object.
- **Test pass rate.** 10 suites / 57 tests green after the migration — solid coverage by govern standards.

## 📝 Minor nits (none blocking)

1. **Defensive `as Nominee[] | undefined` on `getAllRemovedNominees`.** Solidity arrays never return undefined — they return `[]` if empty. The `| undefined` and the `if (!result) return []` guard are unreachable. Could simplify to `as Nominee[]`. Inherited from old code, not introduced here.
2. **`amount` passed as `string` to uint256 args** (`createLock`, `increaseAmount`, `increaseUnlockTime`). Viem accepts string for uint256, so this works, but the build PR (#397) went the other way and explicitly `BigInt()`-wrapped every uint256 arg. The two PRs land different conventions — worth picking one for future migrations (I'd lean toward the explicit `BigInt()` for self-documentation, but it's purely stylistic).
3. **`ethers.toBigInt(response as bigint | string)`** — `response` from viem is already `bigint`. The `| string` cast is dead. Could shrink to `ethers.toBigInt(response as bigint)` or drop the ethers dep entirely with `response === 0n`. Out of scope.
4. **Inconsistent return types across writes** (full receipt vs `transactionHash` string). This faithfully preserves prior behavior, which is the right move for the migration. A follow-up to normalize (probably to full receipts — callers can read `.transactionHash` when they need it) would be cleaner but is purely cosmetic. Don't do it in this PR.
5. **Pre-existing `window.console.log(...)` in catch blocks** — CLAUDE.md says only `console.warn` / `console.error` are allowed. Not introduced here. Worth a separate cleanup PR.

## Verdict

LGTM — clean migration, careful caller-compat verification, sensible consolidation of pre-existing inline reads. Approving.

Three down, two to go (bond, contribute, marketplace per the PR description).

rajat2502 (approved):
LGTM. Solid migration — caller return shapes preserved, dead getters cleanly removed.

Same two cross-cutting concerns as PR #395 to verify post-merge:

1. **Safe-via-WalletConnect**: dropping the `provider.getCode(account)` detection from `sendTransaction` means `waitForTransactionReceipt` will hang on Safe txs awaiting co-signers. Govern likely has more multisig users than launch (governance committees), so worth a manual smoke test of create-lock / vote / donate / revoke-power via a Safe-WC wallet.
2. **Silent Safe UX**: the old wrapper notified `'Please submit the transaction in your safe app.'` for contract accounts; new code is silent. Confirm with product this is the intended post-migration UX.

Inconsistency with PR #395: that one pinned `chainId` defensively on `writeContract` + `waitForTransactionReceipt`; this one only pins it via the spread into `simulateContract`. Worth aligning across the stack — re-pinning on `waitForTransactionReceipt(... { hash, chainId: MAINNET })` is cheap insurance against mid-wait chain switches.

Nits (non-blocking):
- `veOlasParams.abi` is the only param block missing `as Abi` — small consistency fix.
- `(await readContract(...)) as Nominee[] | undefined` and the `if (!result) return []` guard are dead — viem's `readContract` throws rather than returning undefined.
- `ethers.toBigInt(response as bigint | string)` at line 209 can be `response as bigint` (or `BigInt(...)`) and the whole `ethers` import dropped.
- `amount: string` for uint256 args works today because callers feed clean decimal strings — tightening to `bigint` would remove a foot-gun.
- `voteForNomineeWeights`'s `account: Address | undefined` could be tightened to `Address` to match the other writes.

mohandast52 (approved):
LGTM — clean and surgical migration. Verified independently:

### What I checked
- **Decoding-side correctness** (the trap that bit PR #397): every read in this PR returns either a single value, a single tuple/struct, or a `tuple[]` of structs — no multi-named-top-level outputs. So the `as { ... }` casts match viem's actual runtime shape:
  - `getAllRemovedNominees` → `tuple[]` struct array → objects with `.account` / `.chainId` ✅
  - `getLastUserPoint` → single struct → object with `.slope` ✅
  - `lockedEnd` → single uint256 → bigint ✅ (the `as number` cast is technically wrong but JS allows number/bigint comparisons, so `nextWeek >= result` works)
  - `allowance` → single uint256 → bigint, wrapped through `ethers.toBigInt` ✅
  - `getService` (via `readContracts`) → single struct → object with `.state` ✅
  - `verifyInstance` (via `readContracts`) → single bool ✅
- **`web3.ts` deletion is safe** — grep confirms only consumer was `requests.ts`, and the barrel re-export was removed in the same commit.
- **Caller return shapes preserved.** `RevokePower.tsx:143` still reads `result?.transactionHash`; `waitForTransactionReceipt` returns a `TransactionReceipt` with that field, so consumer is unchanged. Other callers either `await` without reading or read `txHash` from functions that explicitly return `receipt.transactionHash`.
- **`chainId` pinned on every contract config** — addresses the concern I raised on #395. Govern is mainnet-only, so the literal pin is appropriate.
- **`BigInt(totalAmount)` for the payable `value`** in `depositServiceDonationsETH` is the correct conversion (web3.js auto-coerced strings; viem requires bigint).
- **Tests pass** locally: 10 suites / 57 tests green. Lint clean.

### Cross-cutting concerns worth verifying post-merge (already raised by others)
- Safe-via-WalletConnect: `waitForTransactionReceipt` will hang on Safe txs awaiting co-signers. Worth a manual smoke of create-lock / vote / donate / revoke-power inside a Safe app, since governance committees skew toward multisig users.
- Defense-in-depth `chainId` on `waitForTransactionReceipt` (and `writeContract`) — currently only pinned via the `...veOlasParams` spread on `simulateContract`. Cheap insurance against mid-wait chain switches.

### Minor nits (non-blocking; happy to do these as a follow-up)
- `as Nominee[] | undefined` + `if (!result) return []` guard is dead — viem's `readContract` throws rather than returning undefined.
- `ethers.toBigInt(response as bigint | string)` could be `response as bigint` (or drop ethers entirely with `response === 0n`).
- Convention drift across the migration series: PR #397 explicitly `BigInt()`-wraps every uint256 arg; this PR passes uint256 args as strings. Both work — worth picking one for bond/contribute/marketplace.

Approving. Three down, two to go (bond, contribute, marketplace).

PR #398, opened 2026-05-19
chore(security): bump 17 vulnerable transitives via yarn resolutions

mohandast52:
## Summary

Closes ~17 Dependabot alerts by forcing patched versions of deep transitives via the `resolutions` block. Every bump is to the first upstream-patched version per the corresponding GHSA.

### Existing resolutions updated
| Package | Was | Now | Closes |
|---|---|---|---|
| `ws` | 8.17.1 | **8.20.1** | #361 (GHSA-58qx-3vcg-4xpx) |
| `h3` | 1.15.6 | **1.15.9** | #245 (GHSA-72gr-qfp7-vwhw), #246 (GHSA-4hxc-9384-m385) |
| `serialize-javascript` | 7.0.3 | **7.0.5** | #264 (GHSA-qj8w-gfj5-8c6v) |

### New resolutions added
| Package | Version | Closes | Pulled by |
|---|---|---|---|
| `@babel/runtime` | 7.29.2 | #80 (GHSA-968p-4wvh-cqc8) | many — 31 distinct `^7.x` consumer ranges |
| `cookie` | 0.7.2 | #61 (GHSA-pxg6-pf52-xh8x) | `express`/`set-cookie-parser`/wagmi network layer |
| `express` | 4.20.0 | #52 (GHSA-qw6h-vgh9-j6wx) | `web3.js → web3-providers-http → ...` |
| `follow-redirects` | 1.16.0 | #306 (GHSA-r4q5-vmmm-2653) | `axios` |
| `micromatch` | 4.0.8 | #43 (GHSA-952p-6rrq-rcjv) | nx/jest/babel tooling |
| `min-document` | 2.19.1 | #144 (GHSA-rx8g-88g5-qh64) | `global` (web3 stack browser shim) |
| `nanoid` | 3.3.11 | #69 (GHSA-mwcw-c2x4-8c55) | many (kept on 3.x — v4/5 are ESM-only) |
| `postcss` | 8.5.10 | #323 (GHSA-qx2v-qp2m-jg93) | `styled-components`/`@nx/next` |
| `qs` | 6.14.2 | #164 (GHSA-6rw7-vpxm-498p), #190 (GHSA-w7fw-mjwx-w883) | `axios`/`express` |
| `send` | 0.19.0 | #55 (GHSA-m6fv-jmcg-4jfg) | `serve-static`/`express` |
| `serve-static` | 1.16.0 | #54 (GHSA-cm22-4g7w-348p) | `express` |
[...]

PR #400, opened 2026-05-19
chore(deps): migrate contribute's web3.js code to viem

atepem:
PR #5 of the web3.js -> viem modernization. Contribute had the most scattered surface — 5 contract getters in common-util/Contracts/index.ts plus a local getEstimatedGasLimit wrapper, consumed across 6 caller files. The Login component also constructed a Web3 instance for a window-global that nothing else read.

Deletions
- apps/contribute/common-util/Contracts/index.ts — the web3.js module.
- apps/contribute/common-util/functions/requests.ts — local getEstimatedGasLimit. All 4 callers migrated to simulateContract + writeContract + waitForTransactionReceipt so the helper is dead.

New
- apps/contribute/common-util/Contracts/params.ts — small module exporting per-contract { address, abi, chainId } tuples for the mainnet contracts (delegateContribute, veOlas) and the base contract (contributors), plus a mintNftParams(chainId) helper for the chain-dependent mint NFT contract (mainnet vs goerli ABIs/addresses). chainId is widened to `number` so wagmi's overload typing doesn't reject the literal against the workspace's non-const chains tuple.

Migrated callers (all to wagmi-native @wagmi/core calls)
- components/MembersList/requests.ts — 1 read (votingPower).
- components/VotingPower/requests.ts — 3 reads + 1 write (delegate). delegate previously routed through getEstimatedGasLimit + .send(); now simulateContract + writeContract + waitForTransactionReceipt.
- components/VotingPower/hooks.ts — caller updates: delegator list return is readonly Address[] (viem inference from const ABI), needs spread to mutable for setState; veolas balance return is bigint, needs String() before storing in string-typed state.
- components/Leaderboard/MintNft/utils.ts — 1 write (mint) + 1 read (tokenURI). The new tokenId is parsed from the Transfer event in the receipt logs via viem decodeEventLog (matching launch's pattern).
- components/Staking/requests.ts — 1 write (createAndStake) on base chain with a payable value (CREATE_AND_STAKE_VALUE bigint).
- components/Profile/Staking/requests.ts — 1 write (reStake) on base.
- common-util/api/nftDetails.ts — 3 reads (totalSupply, ownerOf, tokenURI). Loop iterator switched from string ids to native bigint. getLatestMintedNft account parameter widened to accept undefined.

LoginV2 + global typing
- components/Login/LoginV2.tsx: drops `import Web3 from 'web3'` and the `new Web3(modalProvider)` construction. Removes the `window.WEB3_PROVIDER = wProvider` write — verified by grep that nothing in the repo reads it. Keeps `window.MODAL_PROVIDER`, which libs/util-functions/getModalProvider still reads for the legacy gnosis-safe polling path.
- global.d.ts: drops the WEB3_PROVIDER declaration. MODAL_PROVIDER retyped from `Web3` to `unknown` (the original `Web3` annotation was wrong — MODAL_PROVIDER held the raw EIP-1193 provider, not a Web3 instance).

Notes on viem return-type inference
- The mainnet/base ABI modules in common-util/AbiAndAddresses are declared `as const`, so dropping `as Abi` on these lets viem infer exact return types from the ABI. Caught the readonly-Address[] and bigint-not-string mismatches at build time before they reached runtime — same approach as the build migration after the review.
- mintNftContract*.tsx modules are NOT `as const`, so mintNftParams retains the `as Abi` cast and reads still need manual casts. Out of scope to convert those ABI files in this PR.

Eliminates 2 of the remaining 5 source-file web3.js imports (the contribute Contracts/index.ts and the LoginV2 web3 mirror). After this lands: bond (2), marketplace (1).

- yarn nx build contribute: green
- yarn nx test contribute: green (no spec files exist for this app apart from the missing one — Jest's --passWithNoTests default)
- npx eslint on all changed files: clean

## Types of changes

## Supply-chain checklist


mohandast52 (approved):
LGTM. Reviewed against the ABIs and verified the decode shapes — unlike the bond PR, every read here is either a single value or a single array-typed output, and the contract ABIs (`DELEGATE_CONTRIBUTE_ABI`, `VEOLAS_ABI`, contributors) are declared `as const`, so viem infers exact return types and TypeScript would have caught any object/array mismatch at build time. That's exactly what surfaced the two adaptations in [VotingPower/hooks.ts](apps/contribute/components/VotingPower/hooks.ts):
- `getDelegatorList` → `readonly Address[]`, spread to mutable `[...list]` for `setState` ✅
- `getVotes` → `bigint`, `String(result)` before storing in string-typed state ✅

### Verified
- `votingPower` / `getVotes` → single `uint256` → `bigint`; `mapDelegation` → single `address`; `getDelegatorList` → single `address[]`. All single-output, no tuple-vs-object trap.
- Writes (`delegate`, `reStake`, `createAndStake`, OLAS `approve`) → canonical `simulateContract → writeContract → waitForTransactionReceipt`. `createAndStake` correctly wraps `BigInt(socialId)` and uses the payable `value: CREATE_AND_STAKE_VALUE` (now `2n`).
- `nftDetails.ts` — loop iterator switched to native `bigint` (`i = 1n; i <= total; i += 1n`) to match `totalSupply`'s bigint return; reads cast manually because the mint ABIs aren't `as const`. Correct.
- `LoginV2` — dropping the `new Web3(modalProvider)` mirror and `window.WEB3_PROVIDER` is safe (grep confirms no readers); `window.MODAL_PROVIDER` retained for the legacy gnosis-safe path. The `global.d.ts` retype from `Web3` to `unknown` is more honest than the original (it held the raw EIP-1193 provider).
- Lint clean locally.

### One minor nit (non-blocking)
In [MintNft/utils.ts](apps/contribute/components/Leaderboard/MintNft/utils.ts), the Transfer event's `id` is `uint256`, so `decodeEventLog` returns it as a **bigint**:

```js
return get(args, 'id') || (args.tokenId !== undefined ? args.tokenId.toString() : '');
```

The `id` branch returns a raw `bigint` while the function is typed `Promise<string>` (and the `tokenId` fallback branch correctly calls `.toString()`). It works downstream because `pollNftDetails` interpolates it (`\`${id}\``), but `dispatch(setNftDetails({ tokenId: id, ... }))` then stores a bigint in Redux — which can trip RTK's non-serializable-value check. Suggest normalizing for consistency with the old string contract:

```js
const id = get(args, 'id') ?? args.tokenId;
return id !== undefined ? id.toString() : '';
```

Approving — the nit is cosmetic and doesn't block.

rajat2502 (changes requested):
## Review — web3.js → viem migration (contribute)

Reviewed against base `precious/viem-migration`, 12 files, +248/−267.

### Intent

PR #5 of the web3.js→viem modernization, for the contribute app. Deletes the web3.js `common-util/Contracts/index.ts` (5 contract getters) and the local `getEstimatedGasLimit` helper, replacing them with a `Contracts/params.ts` module of `{address, abi, chainId}` tuples, and rewrites all 6 caller files to `@wagmi/core` (`readContract` / `simulateContract` → `writeContract` → `waitForTransactionReceipt`). Also drops the unused `window.WEB3_PROVIDER` Web3 mirror from `LoginV2`.

Verified: no dangling imports of the deleted modules; `getServiceRegistryL2Contract` had zero callers (dead code, cleanly removed); `WEB3_PROVIDER` is read nowhere; `MODAL_PROVIDER` readers all cast via `as any`, so retyping it to `unknown` breaks nothing.

### 🔴 Blocking

**`components/Leaderboard/MintNft/utils.ts` — `mintNft` returns a `bigint` where a `string` is declared.**

The Transfer event's token-id param is `id` (`uint256`, **indexed**) — confirmed in both mint NFT ABIs. viem `decodeEventLog` decodes an indexed `uint256` to a **`bigint`**. The code returns `get(args, 'id')` with no `String()`, so `mintNft` (declared `Promise<string>`) resolves a bigint. It then flows into `dispatch(setNftDetails({ tokenId: id }))`, whose store field `nftDetails.tokenId` is typed `string` (`store/setup.ts:30`). web3's `returnValues.id` was a string — the runtime type regressed.

This is the main path, not an edge case — every successful mint hits it. Symptoms: Redux Toolkit `serializableCheck` warns on the non-serializable bigint, and `ShowBadge tokenId={...}` receives a bigint. Note `getLatestMintedNft` stores a *string* `tokenId`, so the store ends up holding two runtime types for the same field depending on path.

```ts
if (decoded.eventName === 'Transfer') {
  const args = decoded.args as unknown as { id?: bigint; tokenId?: bigint };
  const tokenId = args.id ?? args.tokenId;
  return tokenId !== undefined ? String(tokenId) : '';
}
```

The `{ id?: string }` annotation is also wrong, and the `args.tokenId` fallback branch is dead code (both chain ABIs name the param `id`).

### Security

No security-relevant changes — input boundaries reviewed (account flows from the connected wallet, contract addresses are constants), no new injection sinks or auth surface. `simulateContract`-before-`writeContract` is a net safety gain.

### Regression risks (verify)

- **`common-util/api/nftDetails.ts` — `getLatestMintedNft`**: the `ownerOf` loop changed from sequential `await`-in-loop to a promise array + `Promise.all` — now **N parallel** `eth_call`s (N = `totalSupply`). For a many-token contract this can trip RPC rate-limiting (429) and throw where it previously succeeded slowly. Consider viem `multicall` (one request) or chunked batches. The `/* eslint-disable no-await-in-loop */` at line 1 is now stale.
- **Write chainId enforcement**: `delegate` (mainnet), `createAndStake`/`reStake` (base) and `mint` are now pinned to an explicit `chainId`; wagmi throws `ChainMismatchError` if the wallet is on a different chain — no auto-switch. Previous web3 `.send()` did not enforce this. Confirm the UI guards chain-switching before these four writes, or surfaces the error gracefully (contribute supports both mainnet + base, so the mismatch is reachable).
- **Gas**: `getEstimatedGasLimit`'s `×1.2` buffer + 2M fallback are gone; gas now relies on the node/wallet estimate. Likely fine (wallets re-estimate), but spot-check the heavy `createAndStake` on base completes.

### Verdict

Requesting changes for the `mintNft` bigint fix. Once that's in and the chain-switch question is answered, this is a clean, well-scoped migration.

Tanya-atatakai (changes requested):
# PR Review: `chore(deps): migrate contribute's web3.js code to viem`

The migration code itself is excellent — arguably the cleanest of the four-PR series so far (requireChainId-style guard, `as const` ABI inference where available, defensive null checks on the chain-dependent mint NFT contract, careful Transfer event parsing for cross-chain ABI shape differences).

**But this PR is unsafe to merge as-is** because the branch silently reverts merged security work in `package.json`. Requesting changes for that reason, not for the migration code.

## 🚨 Blocker — stale-rebase silently reverts PR #398's security resolutions

Running `git diff origin/main..origin/precious/web3-to-viem-contribute -- package.json`:

```diff
@@ -146 +146 @@
-    "h3": "1.15.9",
+    "h3": "1.15.6",
@@ -155 +155 @@
-    "serialize-javascript": "7.0.5",
+    "serialize-javascript": "7.0.3",
@@ -164 +164 @@
-    "ws": "8.20.1",
+    "ws": "8.17.1",
-    "@babel/runtime": "7.29.2",
-    "cookie": "0.7.2",
-    "express": "4.20.0",
-    "follow-redirects": "1.16.0",
-    "micromatch": "4.0.8",
-    "min-document": "2.19.1",
-    "nanoid": "3.3.11",
-    "postcss": "8.5.10",
-    "qs": "6.14.2",
-    "send": "0.19.0",
-    "serve-static": "1.16.0",
-    "tmp": "0.2.4",
-    "webpack": "5.104.1"
```

The `-` lines are what's on `main`; the `+` lines are what this branch has. Translation: merging this PR rolls back PR #398 entirely — 13 security resolutions deleted, 3 patched versions downgraded. Every advisory PR #398 closed reopens, and the `ws@8.17.1` regression specifically is what PR #128 (townhall-kpis) had to patch in the sister repo for GHSA-58qx-3vcg-4xpx.

The PR body's "rebased onto current umbrella" line suggests this branch is stacked on a feature-umbrella branch that hasn't itself merged #398, so the rebase didn't pick those up. Either way, a `git rebase origin/main` on this branch should drop those `-`/`+` lines cleanly (the package.json/yarn.lock changes aren't part of contribute's migration scope).

**Action needed before merge:** rebase onto current `main`, re-run `yarn install`, confirm `package.json`/`yarn.lock` no longer show the diff above. Once that's done I'll re-review and approve.

## ✅ The migration code itself — LGTM

I want to be clear that the actual contribute migration is high-quality work and the only reason this isn't an approval. Things I verified:

- **`params.ts` extraction.** Per-contract `{ address, abi, chainId }` consolidation, mainnet-only constants for `delegateContribute` + `veOlas` + `contributors`, `mintNftParams(chainId)` factory for the chain-dependent contract. `chainId: number` widening with the same inline rationale as launch/build/govern.
- **`mintNft` Transfer event parsing.** Mirrors launch's `InstanceCreated` pattern — `for…of receipt.logs` + checksum-equal address gate + `decodeEventLog` inside a `try/catch` for "log emitted by same address but not in the ABI." The `get(args, 'id') || args.tokenId.toString()` handles the cross-chain ABI key-shape difference (older Mint NFT ABIs used `id`, newer ones use `tokenId`) which is a real footgun if you didn't know about it.
- **`mintNftParams` null guard.** `if (!params) throw new Error('Mint NFT contract not deployed on this chain')` is the right defensive shape for a chain-dependent contract — explicit and surfaces the misconfiguration at the call site instead of letting `undefined.address` fail deep in viem.
- **`as const` ABI inference win.** Dropping `as Abi` on the mainnet/base modules lets viem infer return types from the const ABI, which caught the readonly-Address[] (needs spread to mutable for setState in `VotingPower/hooks.ts`) and bigint-not-string (needs String() before storing in string-typed state) mismatches at build time. This is exactly the kind of TypeScript-as-safety-net argument that the build PR (#397) review series has been making — payoff visible here.
- **LoginV2 cleanup.** Drops the dead `import Web3 from 'web3'` + `new Web3(modalProvider)` + `window.WEB3_PROVIDER = wProvider`. The `global.d.ts` retyping of `MODAL_PROVIDER` from `Web3` to `unknown` (since it actually holds the raw EIP-1193 provider) is the kind of opportunistic correctness fix I like to see in dep-removal sweeps.
- **`requests.ts` deletion.** Local `getEstimatedGasLimit` had 4 callers, all migrated to canonical `simulateContract → writeContract → waitForTransactionReceipt`. Helper is genuinely dead and the deletion is safe.
- **Tests/build.** PR body claims green; consistent with the prior PRs.

## 📝 Minor nits (defer until the rebase is done)

1. **`as unknown as { id?: string; tokenId?: bigint }`** double-cast on the decoded Transfer args. Same pattern as launch — `as const` on the mint NFT ABIs would let viem return typed `args` directly. PR body acknowledges the ABI files aren't `as const` and that's out-of-scope here. Fine to leave.
2. **`window.console.log('Error occurred on minting NFT')` in the `mintNft` catch** is a CLAUDE.md-style violation (only `console.warn`/`console.error` allowed). Pre-existing, not introduced here.
3. **`pollNftDetails` interval has no clearInterval on the error path.** If `fetch` throws *after* the interval fires (network error during polling), the `reject(error)` runs but the interval keeps firing. Pre-existing, not introduced here. Worth a follow-up.

## Verdict

**Request changes** — rebase onto current `main` so the package.json/yarn.lock diff drops the security-resolution reversal. The migration itself is approve-ready; happy to re-approve as soon as the rebase lands and the package.json diff against `main` shows nothing.

rajat2502 (approved):

mohandast52 (approved):
Re-reviewed the updated head (`51101952b`). The mint-`id` nit is resolved and improved beyond what I suggested:

```js
const args = decoded.args as unknown as { id: bigint };
return String(args.id);
```

`String(args.id)` converts the `uint256` bigint to a string at the boundary, matching the store typing (`nftDetails.tokenId: string`), with a comment pointing at `store/setup.ts`. Two additional improvements I didn't ask for but like:
- Address comparison switched from `.toLowerCase()` to viem's checksummed `getAddress()` — more robust.
- `console.error` instead of `window.console.log`, per the repo's console rules.

The rest of the migration is unchanged from my earlier approval — all reads single-output against `as const` ABIs (TypeScript-guarded), writes canonical. Lint clean.

Re-approving the current head.

Tanya-atatakai (approved):
## Re-review at head `51101952b`

Both prior blockers are resolved cleanly.

### ✅ `mintNft` bigint → string (rajat2502's blocker)
`Leaderboard/MintNft/utils.ts:55` now returns `String(args.id)` — the `uint256`-decoded bigint is converted at the boundary before flowing into `dispatch(setNftDetails({ tokenId }))`, matching the store's `nftDetails.tokenId: string` typing. The fallback-to-`tokenId` branch is gone — both ABIs declare `Transfer(from, to, id)`, so the single-shape cast is correct. Bonus improvements I didn't ask for:
- Address comparison switched from `toLowerCase()` to viem's checksummed `getAddress()` — more robust.
- `console.error` instead of `window.console.log` — aligns with the project's console rule.

### ✅ N parallel `ownerOf` calls (rajat2502's regression risk)
`common-util/api/nftDetails.ts` now batches the `ownerOf` reads through `multicall(wagmiConfig, { contracts: calls, allowFailure: true })` — one `eth_call` instead of `totalSupply`-many parallel reads. The `'' as Address` placeholder for failed positions degrades to a clean "not found" via the subsequent `toLower(e) === toLower(account)` check. The stale `eslint-disable no-await-in-loop` is removed.

### ⚠️ Lingering (non-blocking) — 4 security resolutions still missing vs `main`
Tanya-atatakai's earlier blocker (the rollback of 16 #398 resolutions) is mostly resolved — `h3`, `ws`, `serialize-javascript`, and the bulk of the `#398` additions are intact. But `git diff origin/main..origin/precious/web3-to-viem-contribute -- package.json` still shows 4 resolutions absent on the branch:

```diff
-    "bn.js": "5.2.3",
-    "ip-address": "10.1.1",
-    "@web3auth/no-modal/@metamask/sdk": "0.33.1",
-    "@web3auth/no-modal/@metamask/sdk/@metamask/sdk-communication-layer": "0.33.1"
```

These were added to `main` in commits `be812de8` and `1b001384` (security bumps for `bn.js`/`ip-address`/`@metamask/sdk`). The diff against the PR's actual base (`precious/viem-migration`) is empty, so this is strictly a base-branch (umbrella) issue, not something this PR introduced. Whoever merges the `viem-migration` umbrella into `main` needs to take `main`'s resolutions for these four, otherwise the security bumps get silently undone. Flagging on the umbrella rather than blocking this one.

### Minor (non-blocking)
- `delegate` in `VotingPower/requests.ts:30` doesn't pin `chainId` on `waitForTransactionReceipt`, whereas `reStake`/`createAndStake` do (`chainId: base.id`). Likely fine — `writeContract` will throw `ChainMismatchError` first and mainnet is the wagmiConfig default — but inconsistency is worth flattening for grep-ability.
- `mintNftParams(chainId)` still branches on `chainId === 5` (Goerli sunset Jan 2024). Dead branch, pre-existing — out of scope here.
- `pollNftDetails` interval still has no `clearInterval` on the rejection path. Pre-existing.

### Verdict
Approving. The migration is the cleanest of the series — `as const` ABI inference catching the readonly-array and bigint-not-string mismatches at build time is exactly the kind of correctness payoff that justifies the migration. The remaining umbrella-level resolution drift should be tracked on the `viem-migration` PR, not this one.

PR #202, opened 2025-08-22
(marketplace) chore: internal redirects

rajat2502:
## Proposed changes

Added internal redirects for backward compatability with the old apps.

## Types of changes

What types of changes does your code introduce?
_Put an `x` in the boxes that apply_

- [ ] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

PR #401, opened 2026-05-19
chore(security): bump bn.js, ip-address, @metamask/sdk via resolutions

mohandast52:
## Summary

Phase 2 of the alert clear-out. Closes 5 more alerts by forcing patched versions of deep transitives via `resolutions` block.

Combined with the dismissals planned right after this PR merges, target is **0 open alerts**.

## Bumps

| Resolution | Closes | Notes |
|---|---|---|
| `bn.js: 5.2.3` | **#201** (GHSA-378v-28hj-76wf, >=5.0.0, <5.2.3) + **#202** (same advisory, <4.12.3) | v5 is API-compatible with v4 for the consumers in this tree (web3.js, ethers, solana, ethereum-cryptography). Forcing 5.2.3 globally pulls both vulnerable lines onto the patched version. |
| `ip-address: 10.1.1` | **#325** (GHSA-v2v4-37r5-5v8g) | Major bump from 9.0.5. Only consumer is `socks@^2.6.1` (proxy parsing); the 10.x API surface socks uses (`Address4`/`Address6` constructors + `toString`) is back-compatible. |
| `@metamask/sdk: 0.33.1` | **#137** (GHSA-qj3p-xc97-xw74, malicious-debug transitive) | wagmi 2.5.20 pins `^0.33.0` — patch bump is API-compatible. |
| `@metamask/sdk-communication-layer: 0.33.1` | **#138** (same advisory) | Same. |

## Remaining alerts after this PR (to be handled via dismissal)

| Alert | Approach |
|---|---|
| `ajv` #199, #200 — ReDoS via `$data` | Dismiss as `not_used` — `$data` keyword is not used in any schema definitions in our tree (we don't write JSON-Schema). |
| `yaml` #250, #251 — stack overflow on deeply nested YAML | Dismiss as `not_used` — only parses build-time configs (cosmiconfig, postcss-load-config); inputs are author-controlled. |
| `js-yaml` #152, #153 — prototype pollution on merge | Dismiss as `not_used` — dev-scope; only parses build-time configs (ESLint, Nx, @yarnpkg/parsers). |
| `tough-cookie` #16 — prototype pollution | Dismiss as `not_used` — the patched 4.1.3 version is already in the tree (axios chain); the vulnerable 2.5.0 is only pulled in by jsdom@22 (dev test environment, never runs in production). |
| `request` #15 — SSRF in deprecated package | Dismiss as `not_used` — only consumer is `servify` (web3@1's deprecated provider util) which isn't called in any of our apps. |
| `web3-core-subscriptions` #139 — prototype pollution | Dismiss as `not_used` — only triggered if we use web3.js event subscriptions; verified no `subscribe(...)` calls in code. |
| `@stablelib/ed25519` #270 — signature malleability | Dismiss as `tolerable_risk` — no upstream patch; advisory affects signature malleability which is not exploitable in our chain-of-trust verification (we don't accept user-submitted signatures in any signing path). |
| `elliptic` #167 — risky-impl crypto primitive | Dismiss as `tolerable_risk` — already at latest 6.6.1 (no patched version exists); the risky primitive isn't used in our wagmi/ethers code paths. |
| `@tootallnate/once` #223 — incorrect control flow | Dismiss as `tolerable_risk` — dev-scope, low severity, can't bump to 3.0.1 due to ESM/CJS boundary issue documented in #398 review. |

## Test plan

Verified locally on Node 22.18.0:

- [x] `yarn install` — succeeds; yarn.lock 514 lines lighter (dedup wins from forcing v5 bn.js)
- [x] `yarn supply-chain` — all 3 gates pass
- [x] `yarn nx test {build,contribute,govern,launch}` — pass
- [x] `yarn nx run-many --target=build --projects=<all 9 apps>` — all 9 build successfully

CI will repeat all checks; Vercel previews will land for manual smoke test.

### DIFF
--- a/package.json
+++ b/package.json
@@ -179,6 +179,10 @@
     "send": "0.19.0",
     "serve-static": "1.16.0",
     "tmp": "0.2.4",
-    "webpack": "5.104.1"
+    "webpack": "5.104.1",
+    "bn.js": "5.2.3",
+    "ip-address": "10.1.1",
+    "@metamask/sdk": "0.33.1",
+    "@metamask/sdk-communication-layer": "0.33.1"
   }
 }
--- a/package.json
+++ b/package.json
@@ -182,7 +182,7 @@
     "webpack": "5.104.1",
     "bn.js": "5.2.3",
     "ip-address": "10.1.1",
-    "@metamask/sdk": "0.33.1",
-    "@metamask/sdk-communication-layer": "0.33.1"
+    "@web3auth/no-modal/@metamask/sdk": "0.33.1",
+    "@web3auth/no-modal/@metamask/sdk/@metamask/sdk-communication-layer": "0.33.1"
   }
 }
--- a/yarn.lock
+++ b/yarn.lock
@@ -342,7 +342,7 @@
   dependencies:
     "@babel/types" "^7.23.0"
 
-"@babel/helper-module-imports@^7.0.0", "@babel/helper-module-imports@^7.16.7", "@babel/helper-module-imports@^7.22.15", "@babel/helper-module-imports@^7.24.1", "@babel/helper-module-imports@^7.24.3":
+"@babel/helper-module-imports@^7.0.0", "@babel/helper-module-imports@^7.22.15", "@babel/helper-module-imports@^7.24.1", "@babel/helper-module-imports@^7.24.3":
   version "7.24.3"
   resolved "https://registry.yarnpkg.com/@babel/helper-module-imports/-/helper-module-imports-7.24.3.tgz#6ac476e6d168c7c23ff3ba3cf4f7841d46ac8128"
   integrity sha512-viKb0F9f2s0BCS22QSF308z/+1YWKV/76mwt61NBzS5izMzDPwdq1pTrzf+Li3npBWX9KdQbkeCt1jSAM7lZqg==
@@ -1272,7 +1272,7 @@
   resolved "https://registry.yarnpkg.com/@babel/regjsgen/-/regjsgen-0.8.0.tgz#f0ba69b075e1f05fb2825b7fad991e7adbb18310"
   integrity sha512-x/rqGMdzj+fWZvCOYForTghzbtqPDZ5gPwaoNGHdgDfF2QA/XZbCBp4Moo5scrkAMPhB7z26XM/AaHuIJdgauA==
 
-"@babel/runtime@7.29.2", "@babel/runtime@^7.10.1", "@babel/runtime@^7.10.4", "@babel/runtime@^7.11.1", "@babel/runtime@^7.11.2", "@babel/runtime@^7.12.5", "@babel/runtime@^7.16.7", "@babel/runtime@^7.17.2", "@babel/runtime@^7.17.9", "@babel/runtime@^7.18.0", "@babel/runtime@^7.18.3", "@babel/runtime@^7.20.0", "@babel/runtime@^7.20.6", "@babel/runtime@^7.20.7", "@babel/runtime@^7.21.0", "@babel/runtime@^7.22.5", "@babel/runtime@^7.22.6", "@babel/runtime@^7.23.2", "@babel/runtime@^7.23.4", "@babel/runtime@^7.23.6", "@babel/runtime@^7.23.9", "@babel/runtime@^7.24.4", "@babel/runtime@^7.24.7", "@babel/runtime@^7.24.8", "@babel/runtime@^7.25.0", "@babel/runtime@^7.25.7", "@babel/runtime@^7.26.0", "@babel/runtime@^7.26.9", "@babel/runtime@^7.27.6", "@babel/runtime@^7.7.2", "@babel/runtime@^7.8.4", "@babel/runtime@^7.9.2":
+"@babel/runtime@7.29.2", "@babel/runtime@^7.10.1", "@babel/runtime@^7.10.4", "@babel/runtime@^7.11.1", "@babel/runtime@^7.11.2", "@babel/runtime@^7.12.5", "@babel/runtime@^7.16.7", "@babel/runtime@^7.17.2", "@babel/runtime@^7.17.9", "@babel/runtime@^7.18.0", "@babel/runtime@^7.18.3", "@babel/runtime@^7.20.0", "@babel/runtime@^7.20.7", "@babel/runtime@^7.21.0", "@babel/runtime@^7.22.5", "@babel/runtime@^7.22.6", "@babel/runtime@^7.23.2", "@babel/runtime@^7.23.4", "@babel/runtime@^7.23.6", "@babel/runtime@^7.23.9", "@babel/runtime@^7.24.4", "@babel/runtime@^7.24.7", "@babel/runtime@^7.24.8", "@babel/runtime@^7.25.0", "@babel/runtime@^7.25.7", "@babel/runtime@^7.26.0", "@babel/runtime@^7.26.9", "@babel/runtime@^7.27.6", "@babel/runtime@^7.7.2", "@babel/runtime@^7.8.4", "@babel/runtime@^7.9.2":
   version "7.29.2"
   resolved "https://registry.yarnpkg.com/@babel/runtime/-/runtime-7.29.2.tgz#9a6e2d05f4b6692e1801cd4fb176ad823930ed5e"
   integrity sha512-JiDShH45zKHWyGe4ZNVRrCjBz8Nh9TMmZG1kh4QTK8hCBTWBi8Da+i7s1fJw7/lYpM4ccepSNfqzZ/QvABBi5g==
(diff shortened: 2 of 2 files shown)


rajat2502 (dismissed):

Tanya-atatakai (commented):
# PR Review: `chore(security): bump bn.js, ip-address, @metamask/sdk via resolutions`

Phase 2 of the burn-down — tight scope (4 lines in `package.json`, 514 deletions in `yarn.lock` from the bn.js dedup), CI/Vercel previews green across all 9 apps, and the dismissal plan for the remaining alerts is well-reasoned with a clean `not_used` vs `tolerable_risk` split.

Leaving as **COMMENTED** rather than approving because the `@metamask/sdk` bump description doesn't match the actual consumer chain — that mismatch matters and deserves a closer look before merge.

## ✅ Clean bumps

- **`bn.js: 5.2.3`** — verified the entry collapses ALL ranges (`bn.js@4.11.6` exact pin, `^4.0.0`, `^4.1.0`, `^4.11.6`, `^4.11.8`, `^4.11.9`, `^5.0.0`, `^5.1.1`, `^5.1.2`, `^5.2.0`, `^5.2.1`, `^5.2.2`) into a single 5.2.3 entry. The v4 → v5 jump for the exact `4.11.6` pin is the kind of thing I'd normally flag, but bn.js has maintained good API stability since 2017 across v4/v5 and the listed consumers (web3.js, ethers, solana, ethereum-cryptography) all work with either major. The 514-line lockfile shrinkage is exactly the dedup win you'd expect.

- **`ip-address: 10.1.1`** — major bump 9.0.5 → 10.1.1 with documented consumer surface (socks `Address4`/`Address6` constructors + `toString`). ip-address v10 is a TypeScript rewrite but kept the named-export constructors and `toString` intact, so the claim holds. Worth a one-time grep through `node_modules/socks` source for any legacy method calls (`is4to6`, `correctForm`, etc.) just to confirm, but I trust the analysis.

- **`@metamask/sdk-communication-layer: 0.33.1`** — clean single-resolution bump.

## ⚠️ The `@metamask/sdk` bump description doesn't match the actual chain

The PR says:
> wagmi 2.5.20 pins `^0.33.0` — patch bump is API-compatible.

But the actual consumer chain from `yarn.lock` is:

```
wagmi@2.5.20
└─ @wagmi/connectors@4.1.26
   └─ "@metamask/sdk" "0.14.3"   ← EXACT pin, not ^0.33.0
```

The `^0.33.0` consumer is a different package: `@web3auth/no-modal@10.3.0`. So the actual lockfile entry after this PR is:

```
"@metamask/sdk@0.14.3", "@metamask/sdk@0.33.1", "@metamask/sdk@^0.33.0":
  version "0.33.1"
```

The `^0.33.0` → `0.33.1` part is genuinely a patch bump (low risk). But the `0.14.3` → `0.33.1` part is a **~19-minor-version API jump** for `@wagmi/connectors@4.1.26`'s bundled MetaMask connector. That connector was authored against the 0.14.x API surface and is now running against 0.33.x:

- 0.14 → 0.33 spans a substantial rewrite of `@metamask/sdk` — constructor signatures, provider event names, connection lifecycle, and re-connect-on-page-load handling all evolved across that range.
- The basic `connect()` / `getProvider()` / `disconnect()` path *might* still work because @metamask/sdk has tried to maintain top-level shape, but edge cases (mobile deep-link, account-switch events, reconnect-on-refresh) are exactly where API drift surfaces.

**The CI signal doesn't cover this.** TypeScript compile + Vercel build only verify the JS bundles. The MetaMask wallet flow only fires when a user clicks "Connect MetaMask" in the browser — which neither Jest nor `next build` exercises.

### What I'd want to see before merging

Either:

1. **Manual smoke on a Vercel preview** — connect via MetaMask (extension), disconnect, reconnect after a page refresh, switch accounts. Repeat for at least one app. Two would be ideal (e.g., `bond` and `govern` since they're the highest-traffic web3 surfaces). If MetaMask Mobile is in scope, deep-link from the QR connect flow.

2. **Or: scope the resolution** so wagmi connectors stay on the 0.14.x line while @web3auth gets the bump:

   ```json
   "@wagmi/connectors/@metamask/sdk": "0.14.3",
   "@wagmi/connectors/@metamask/sdk-communication-layer": "0.14.3",
   "@metamask/sdk": "0.33.1",
   "@metamask/sdk-communication-layer": "0.33.1"
   ```

   This means alert #137/#138 stay open on the 0.14.3 path. The advisory is `GHSA-qj3p-xc97-xw74` (malicious-debug transitive). If the 0.14.3 path's `debug` dep is itself a patched version, the advisory is moot for that subtree — worth checking before deciding which path is the right one. If 0.14.3 still pulls the vulnerable `debug`, we have a real tension between "patch the alert" and "don't break the wagmi connector" and that needs an explicit decision.

Note that the same pattern from #398 review applies here: bare-keyed resolutions in Yarn 1.x apply globally and will silently override exact pins. The wagmi-connectors situation is precisely the kind of "silent override of an exact pin across a major API boundary" that I called out on `@tootallnate/once` last time.

## 📝 Smaller observations

- **Update the PR description's `@metamask/sdk` row** with the actual consumer chain. The current wording implies only `^0.33.0` consumers exist; that's incorrect and reviewers (incl. future-you) will trust it.
- **Dismissal plan is well-reasoned.** The `not_used` vs `tolerable_risk` split is exactly right; `@tootallnate/once` correctly deferred per #398. `@stablelib/ed25519` and `elliptic` reasoning (chain-of-trust, no user-submitted sigs) is the kind of thing worth keeping a one-line note about in `SUPPLY-CHAIN-SECURITY.md` so it doesn't need re-derivation in 6 months.
- **`bn.js@4.11.6` exact pin override is documented in the PR** as part of the v4→v5 claim, but worth calling out explicitly in the lockfile commentary: it's an exact pin being silently bumped, just like the @metamask/sdk case but at a less risky API boundary.

## Verdict

Three of four bumps look great. The fourth's description is wrong about the consumer chain and the actual change (0.14.3 → 0.33.1 for wagmi connectors) is a real runtime risk that build/CI signal can't catch. Happy to approve once either (a) a Vercel-preview MetaMask connect flow is smoke-tested, or (b) the resolution is scoped to leave wagmi's bundled MetaMask connector on 0.14.x.

mohandast52:
Thanks @Tanya-atatakai — sharp catch, and the consumer-chain analysis was exactly right. Fixed in 1b0013844.

**Confirmed your finding + one more detail that simplifies it:**

The advisory (GHSA-qj3p-xc97-xw74) vulnerable range is `>= 0.16.0, <= 0.33.0`. wagmi's `@metamask/sdk@0.14.3` is **below** that range — it was never vulnerable. Only `@web3auth/no-modal`'s `^0.33.0` (→ 0.33.0) was in range. So the bare bump was forcing a risky 0.14.3 → 0.33.1 jump on the wagmi connector for **no security benefit** (wagmi's path didn't need patching).

I also checked the actual root cause: the malicious `debug@4.4.2` is **not in our tree** (we resolve `debug` to 4.4.1/4.4.3). Dependabot flags #137/#138 purely on the `@metamask/sdk` version range, not the real debug version.

**On your scoped-resolution snippet:** I tried the bare + `@wagmi/connectors/@metamask/sdk: 0.14.3` combo first — but in Yarn 1.x the **bare key overrides the path-scoped one** (verified empirically: wagmi still got forced to 0.33.1). Same Yarn-1.x-quirk family as the bare-key behavior you flagged on #398. So I inverted it: dropped the bare resolutions entirely and scoped **only the vulnerable web3auth path**:

\`\`\`json
"@web3auth/no-modal/@metamask/sdk": "0.33.1",
"@web3auth/no-modal/@metamask/sdk/@metamask/sdk-communication-layer": "0.33.1"
\`\`\`

Resulting lockfile:

| Path | Version | Status |
|---|---|---|
| `@metamask/sdk@0.14.3` (wagmi) | 0.14.3 | untouched — byte-for-byte main |
| `@metamask/sdk@^0.33.0` (web3auth) | 0.33.1 | patched |
| `@metamask/sdk-communication-layer@0.14.3` (wagmi) | 0.14.3 | untouched |
| `@metamask/sdk-communication-layer` (web3auth) | 0.33.1 | patched |

Vulnerable 0.33.0 is gone (closes #137/#138) and **wagmi's MetaMask connector is identical to main** — zero runtime risk, so no MetaMask smoke-test gate needed.

Re your smaller notes:
- **PR description updated** with the correct consumer chain.
- **bn.js@4.11.6 exact-pin → 5.2.3** — agreed it's a silent exact-pin bump; called it out explicitly in the description now. bn.js v4/v5 API stability holds for our consumers (web3/ethers/solana).
- **ip-address socks grep** — checked `node_modules/socks`: only uses `Address4`/`Address6` + `toString`, no legacy `is4to6`/`correctForm`. Holds.
- **SUPPLY-CHAIN-SECURITY.md note** for the @stablelib/ed25519 + elliptic chain-of-trust reasoning — will add in the dismissal follow-up.

Re-verified: supply-chain 3/3, tests {build,contribute,govern,launch}, build all 9 apps. CI re-running now.

rajat2502 (approved):

atepem (approved):

PR #402, opened 2026-05-19
chore(deps): migrate bond's web3.js code to viem

atepem:
PR #6 of the web3.js -> viem modernization. Bond was the second-heaviest app to migrate — 4 caller files, 7 contract getters in common-util/functions/web3.js, and a Login web3-mirror that was dead code (same pattern as contribute had).

apps/bond/common-util/functions/web3.js
- Reduced to a single export: getUniswapV2PairContractByChain, which was already viem-based (returns a viem contract bound to a public client for cross-chain LP price reads).
- Removed: getWeb3Details, getContract (internal), getDepositoryContract, getTokenomicsContract, getUniswapV2PairContract, getErc20Contract, getGenericBondCalculatorContract (had zero callers — dead code).

apps/bond/common-util/Contracts/params.js (new)
- Per-contract `{ address, abi, chainId }` helpers: depositoryParams(chainId), tokenomicsParams(chainId), bondCalculatorParams(chainId), uniswapV2PairParams(address, chainId), erc20Params(address, chainId). chainId widened to `number` so wagmi's simulate/write/wait overloads accept it against the workspace's non-const chains tuple.

apps/bond/components/BondingProducts/Bonding/BondingList/useBondingList.jsx
- getLastIDFRequest -> readContract(tokenomicsParams, 'getLastIDF').
- getLpTokenDetails: token0/token1 reads now run in parallel via Promise.all + readContract on uniswapV2PairParams; symbol read via erc20Params. Net 3 sequential awaits -> 1 Promise.all + 1 await.
- useProductListRequest: getProducts read via readContract(depositoryParams).

apps/bond/components/BondingProducts/Bonding/Deposit/useDeposit.jsx
- hasSufficientTokenRequest / getLpBalanceRequest: readContract reads. hasSufficientToken now uses bigint comparison via BigInt(response) instead of relying on web3.js's string coercion.
- approveRequest / depositRequest: simulateContract + writeContract + waitForTransactionReceipt. Drops the local sendTransaction + getEstimatedGasLimit dance.

apps/bond/components/BondingProducts/requests.js + MyBonds/requests.js
- Near-identical files (cross-file duplication is pre-existing); both migrated to the same shape.
- mapUserBonds / getBonds / getBondStatus reads -> readContract loops via Promise.all (was sequential .call() pushes onto an array).
- redeemRequest -> simulateContract + writeContract + waitForTransactionReceipt.
- epoch reads (epochCounter, mapEpochTokenomics, epochLen) -> readContract; bigint -> Number() at the boundary as needed. mapEpochTokenomics epoch arg now wrapped in BigInt() (uint256).

apps/bond/common-util/Login/LoginV2.jsx
- Drops `import Web3 from 'web3'` and `const wProvider = new Web3(...)`.
- Removes `window.WEB3_PROVIDER = wProvider` — verified by repo-wide grep that nothing reads it. Keeps `window.MODAL_PROVIDER`, which libs/util-functions/getModalProvider still reads for the legacy gnosis-safe polling path.

Eliminates the last 2 of 9 source-file web3.js imports across apps/bond/. After this lands, only marketplace remains (1 file). The `web3` direct dep can be considered for removal once marketplace is migrated.

- yarn nx build bond: green (213s)
- npx eslint on all changed files: 0 errors, 3 pre-existing react-hooks/exhaustive-deps warnings in LoginV2.jsx (unrelated)
- yarn nx test bond fails on a pre-existing nx config bug (project.json references apps/bond/jest.config.ts but the actual file is jest.config.js) — unrelated to this PR. Same single existing spec (apps/bond/specs/index.spec.tsx) was already failing on main.

## Types of changes

## Supply-chain checklist

### DIFF
--- a/apps/bond/common-util/Contracts/params.js
+++ b/apps/bond/common-util/Contracts/params.js
@@ -1,0 +1,42 @@
+import {
+  BOND_CALCULATOR,
+  DEPOSITORY,
+  ERC20_ABI,
+  TOKENOMICS,
+  UNISWAP_V2_PAIR_ABI,
+} from 'libs/util-contracts/src/lib/abiAndAddresses';
+
+import { ADDRESSES } from 'common-util/constants/addresses';
+
+// chainId is widened to `number` so wagmi's simulate/write/wait overloads
+// accept it against the workspace's non-const chains tuple.
+
+export const depositoryParams = (chainId) => ({
+  address: ADDRESSES[chainId].depository,
+  abi: DEPOSITORY.abi,
+  chainId: Number(chainId),
+});
+
+export const tokenomicsParams = (chainId) => ({
+  address: ADDRESSES[chainId].tokenomics,
+  abi: TOKENOMICS.abi,
+  chainId: Number(chainId),
+});
+
+export const bondCalculatorParams = (chainId) => ({
+  address: ADDRESSES[chainId].genericBondCalculator,
+  abi: BOND_CALCULATOR.abi,
+  chainId: Number(chainId),
+});
+
+export const uniswapV2PairParams = (address, chainId) => ({
+  address,
+  abi: UNISWAP_V2_PAIR_ABI,
+  chainId: Number(chainId),
+});
+
+export const erc20Params = (address, chainId) => ({
+  address,
+  abi: ERC20_ABI,
+  chainId: Number(chainId),
+});
--- a/apps/bond/common-util/Contracts/params.js
+++ b/apps/bond/common-util/Contracts/params.js
@@ -1,5 +1,4 @@
 import {
-  BOND_CALCULATOR,
   DEPOSITORY,
   ERC20_ABI,
   TOKENOMICS,
@@ -20,12 +19,6 @@ export const depositoryParams = (chainId) => ({
 export const tokenomicsParams = (chainId) => ({
   address: ADDRESSES[chainId].tokenomics,
   abi: TOKENOMICS.abi,
-  chainId: Number(chainId),
-});
-
-export const bondCalculatorParams = (chainId) => ({
-  address: ADDRESSES[chainId].genericBondCalculator,
-  abi: BOND_CALCULATOR.abi,
   chainId: Number(chainId),
 });
 
--- a/apps/bond/common-util/Login/LoginV2.jsx
+++ b/apps/bond/common-util/Login/LoginV2.jsx
@@ -4,7 +4,6 @@ import { useEffect } from 'react';
 import { useDispatch } from 'react-redux';
 import styled from 'styled-components';
 import { useAccount, useBalance, useDisconnect } from 'wagmi';
-import Web3 from 'web3';
 
 import { CannotConnectAddressOfacError } from 'libs/ui-components/src';
 import { MEDIA_QUERY } from 'libs/ui-theme/src';
@@ -82,16 +81,10 @@ export const LoginV2 = ({ onConnect: onConnectCb, onDisconnect: onDisconnectCb }
           connector?.options?.getProvider?.() || (await connector?.getProvider?.());
 
         if (modalProvider) {
-          // We plug the initial `provider` and get back
-          // a Web3Provider. This will add on methods and
-          // event listeners such as `.on()` will be different.
-          const wProvider = new Web3(modalProvider);
-
-          // *******************************************************
-          // ************ setting to the window object! ************
-          // *******************************************************
+          // libs/util-functions/getModalProvider reads window.MODAL_PROVIDER
+          // for the legacy gnosis-safe polling path. The previous web3-instance
+          // mirror (window.WEB3_PROVIDER) had no readers anywhere in the repo.
           window.MODAL_PROVIDER = modalProvider;
-          window.WEB3_PROVIDER = wProvider;
 
           if (modalProvider?.on) {
             // https://docs.ethers.io/v5/concepts/best-practices/#best-practices--network-changes
--- a/apps/bond/common-util/functions/ethers.js
+++ b/apps/bond/common-util/functions/ethers.js
@@ -20,12 +20,3 @@ export const parseToSolDecimals = (amount) => ethers.parseUnits(`${amount}`, 8).
  * divides the amount by 10^8
  */
 export const parseToSol = (amount) => ethers.formatUnits(`${amount}`, 8);
-
-/**
- * TODO: move to autonolas-library and figure out a better way
- * to fetch timestamp
- */
-export const getBlockTimestamp = async (block = 'latest') => {
-  const temp = await window?.WEB3_PROVIDER.eth.getBlock(block);
-  return temp.timestamp * 1;
-};
--- a/apps/bond/common-util/functions/web3.js
+++ b/apps/bond/common-util/functions/web3.js
@@ -1,55 +1,13 @@
 import { createPublicClient, getContract as getViemContract, http } from 'viem';
-import Web3 from 'web3';
 
 import { CHAINS, RPC_URLS } from 'libs/util-constants/src';
-import {
-  BOND_CALCULATOR,
-  DEPOSITORY,
-  ERC20_ABI,
-  TOKENOMICS,
-  UNISWAP_V2_PAIR_ABI,
-} from 'libs/util-contracts/src/lib/abiAndAddresses';
-
-import { ADDRESSES } from 'common-util/constants/addresses';
-import { getChainId, getProvider } from 'common-util/functions/frontend-library';
+import { UNISWAP_V2_PAIR_ABI } from 'libs/util-contracts/src/lib/abiAndAddresses';
 
 /**
- * returns the web3 details
+ * Returns a viem contract instance bound to a public client for the given chain.
+ * Used for cross-chain LP price reads where the connected wallet's chain
+ * doesn't match the LP's chain.
  */
-const getWeb3Details = () => {
-  const chainId = getChainId();
-  const web3 = new Web3(getProvider());
-  return { web3, chainId };
-};
-
-/**
- * returns the contract instance
- * @param {Array} abi - abi of the contract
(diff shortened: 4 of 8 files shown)


mohandast52 (changes requested):
Requesting changes — there are **three runtime-breaking regressions** here, all the same class of bug as PR #397 (multiple top-level contract outputs decode as a viem **tuple/array**, not an object with named keys). Because bond is plain `.js`/`.jsx`, there's no TypeScript safety net, so `yarn nx build bond: green` doesn't catch any of them. I verified each against the actual `DEPOSITORY` ABI by running viem's `decodeFunctionResult` locally.

### 1. `getBonds` → `{ bondIds }` is `undefined` → `.map` crash (hard crash)

`getBonds(account, matured)` has two top-level outputs (`uint256[] bondIds`, `uint256 payout`), so viem returns `[ [1n,2n,3n], 999n ]`. The destructure in both [MyBonds/requests.js:45](apps/bond/components/MyBonds/requests.js#L45) and [BondingProducts/requests.js:50](apps/bond/components/BondingProducts/requests.js#L50):

```js
const { bondIds } = response;        // undefined
const idsList = bondIds.map(...);    // TypeError: Cannot read properties of undefined (reading 'map')
```

This throws on every "My Bonds" / bonding-list load. Fix:
```js
const [bondIds] = response;          // first tuple element
const idsList = bondIds.map((id) => `${id}`);
```

### 2. `getBondStatus` → `...bond` spreads array indices → `.matured` always `undefined`

`getBondStatus(id)` returns `[uint256 payout, bool matured]` → viem gives `[123n, true]`. Spreading that into an object ([requests.js:58-61](apps/bond/components/MyBonds/requests.js#L58-L61)) yields `{ 0: 123n, 1: true, bondId, key }` — there is no `.payout` or `.matured` key. Downstream:
- [requests.js:70](apps/bond/components/MyBonds/requests.js#L70) `.filter((bond) => !bond.matured)` → `!undefined` is always `true`, so matured-bond filtering is silently inverted/broken.
- [MyBonds.jsx:175](apps/bond/components/MyBonds/MyBonds.jsx#L175) `disabled: !record.matured` → always `disabled: true`; bonds can never be redeemed from the UI.

Fix — destructure in the map:
```js
const bondsListWithDetails = allListResponse.map(([payout, matured], index) => ({
  payout,
  matured,
  bondId: idsList[index],
  key: idsList[index],
}));
```

### 3. `mapUserBonds` → `.maturity` is `undefined` → `NaN` maturity dates

`mapUserBonds(bondId)` returns 4 top-level outputs (`address account, uint96 payout, uint32 maturity, uint32 productId`) → viem gives `['0x..', 50n, 1700000000, 7]`. At [MyBonds/requests.js:28](apps/bond/components/MyBonds/requests.js#L28) and [BondingProducts/requests.js:30](apps/bond/components/BondingProducts/requests.js#L30):

```js
maturityDate: Number(lpTokenNameList[index].maturity) * 1000,   // Number(undefined) * 1000 → NaN
```

Fix — index into the tuple (`maturity` is element 2):
```js
maturityDate: Number(lpTokenNameList[index][2]) * 1000,
```

---

### What's correct (checked, no change needed)
- `getProducts` → single `uint256[]` output → viem returns the array directly; `getProductDetailsFromIds(productIdList)` is fine.
- `getLastIDF` → single `uint256` → `Number(lastIdfResponse)` is fine.
- `mapEpochTokenomics` → single `struct EpochPoint` tuple → viem **does** return a named object, so `prevEpochPoint.endTime` works (same as the build/govern migrations). `BigInt(epochNum)` arg wrap is correct.
- `token0` / `token1` / `symbol` / `allowance` / `balanceOf` → all single-output, fine. The `Promise.all` parallelization of `token0`/`token1` is a nice touch.
- All writes (`approve`, `deposit`, `redeem`) → canonical `simulateContract → writeContract → waitForTransactionReceipt`, correct. `BigInt(response) >= BigInt(tokenAmount)` in `hasSufficientTokenRequest` is the right defensive coercion.
- `web3.js` deletion, `params.js` extraction, and the dead `window.WEB3_PROVIDER` removal are all clean.

### Recommendation
The structure of the migration is good and matches the rest of the series — but these three reads need tuple destructuring before merge. Since bond is JS, please add a manual smoke test (or convert these ABIs to `as const` + a typed wrapper) so this class of bug is caught: load **My Bonds** with at least one active and one matured bond, confirm maturity dates render (not "Invalid Date"/NaN), the matured/active filter splits correctly, and the redeem button enables for matured bonds.

The build (#397) hit this exact issue and the fix there was the reference; the difference is TypeScript caught it there and can't here. Happy to re-review promptly once these are addressed.

Tanya-atatakai (approved):
# Re-review on PR #402 — `chore(deps): migrate bond's web3.js code to viem`

mohandast52 ([review here](https://github.com/valory-xyz/autonolas-frontend-mono/pull/402#pullrequestreview)) identified three runtime-breaking regressions of the same class as PR #397: multiple-top-level-output reads decoded as viem tuples but consumed as named objects. Commit `6565e08c` ("fix: use tuple destructuring for multi-output viem readContract calls") was pushed after that review. I verified the fix against the regressions one-by-one:

| Regression | Fix in 6565e08c | Status |
|---|---|---|
| `getBonds` → `{ bondIds }` is `undefined` (2 outputs: `bondIds`, `payout`) | `const [bondIds] = response` in both `MyBonds/requests.js` and `BondingProducts/requests.js` | ✅ Fixed |
| `getBondStatus` → `...bond` spread yields array indices, `.matured` always undefined → matured-filter inverted, redeem button always disabled | `.map(([payout, matured], index) => ({ payout, matured, bondId, key }))` in both files | ✅ Fixed |
| `mapUserBonds` → `.maturity` undefined → `NaN` maturity dates (4 outputs: `account`, `payout`, `maturity`, `productId`) | `lpTokenNameList[index][2]` (positional access for `maturity`) in both files | ✅ Fixed |

The fix is applied symmetrically in both `MyBonds/requests.js` and `BondingProducts/requests.js` (which share near-identical code per the PR description's pre-existing duplication note). Each fix matches mohandast52's prescribed shape exactly.

## ✅ Approving

The structural concerns from mohandast52's review are resolved. The rest of the migration (web3.js deletion, `params.js` extraction, Login web3-mirror cleanup, Promise.all parallelization of `token0`/`token1`, canonical write triplet) matches the pattern of the prior approved PRs in this series.

## 📝 Bot's inline comments — minor cleanups to consider in this or a follow-up

The `valory-coding-agent` bot left four reasonable inline suggestions:

1. **`lpTokenNameList[index][2]` is a magic positional index.** Destructuring at the call site (`const [, , maturity] = lpTokenNameList[index];`) would be more readable and self-documenting; if the `mapUserBonds` ABI shape ever changes (e.g. someone reorders the struct fields), the destructure pattern at least makes the breakage searchable. The current `[2]` works, just sharper-edged. Both `requests.js` files have the same pattern.
2. **`bondCalculatorParams` export with zero callers.** PR body explicitly notes `getGenericBondCalculatorContract` was dead code; the new helper that replaced it is also unused. Consider deleting from `params.js`; re-adding is trivial when a caller arrives.
3. **`window.WEB3_PROVIDER` comment inaccuracy.** The comment claims "no readers anywhere in the repo," but `common-util/functions/ethers.js:29` still references it inside `getBlockTimestamp`. That function has no callers in the bond app (genuinely dead code), so no runtime impact — but the claim isn't quite right. Either tighten the comment ("no live readers" / "only stale reference in dead `getBlockTimestamp`") or delete `getBlockTimestamp` to make the claim honest.

None of these block this PR. Items (2) and (3) are dead-code housekeeping that pairs naturally with the web3.js sunset work; (1) is a readability suggestion. Happy to see them addressed here or in a follow-up.

## Test gap

Bond's jest config is broken on `main` (`project.json` references `jest.config.ts` but the file is `jest.config.js`) — the pre-existing failure means `yarn nx test bond` can't currently exercise these reads. Worth filing as a separate cleanup; without it, this class of "viem-tuple-decoded-as-object" bug will keep being a JS-side footgun. The TypeScript apps (`build`, `govern`, `contribute` — the ABIs declared `as const`) caught this class of bug at compile time. Bond can't.

Approving — fix is verified, the structural concerns are resolved.

rajat2502 (approved):

mohandast52 (approved):
Re-reviewed `6565e08c` + `eda0c5fa`. All three runtime bugs are fixed correctly with positional tuple destructuring — verified the indices against the `DEPOSITORY` ABI:

- **`getBonds`** → `const [bondIds] = response;` ✅ (outputs: `bondIds[0]`, payout[1])
- **`getBondStatus`** → `allListResponse.map(([payout, matured], index) => ({ payout, matured, ... }))` ✅ (outputs: payout[0], matured[1])
- **`mapUserBonds`** → `const [, , maturity] = lpTokenNameList[index];` ✅ (outputs: account[0], payout[1], maturity[2], productId[3])

Both `MyBonds/requests.js` and `BondingProducts/requests.js` fixed identically, with a clear comment explaining viem's positional decoding. Nice side effect: the explicit `payout` field also repairs the `dataIndex: 'payout'` column in [MyBonds.jsx:21](apps/bond/components/MyBonds/MyBonds.jsx#L21), which the old `...bond` spread had silently left under key `0`.

Bonus cleanups in this round are good catches:
- Removed `getBlockTimestamp` from `common-util/functions/ethers.js` — it dereferenced `window.WEB3_PROVIDER`, which this PR deletes, so it would've been a latent broken reference. Confirmed zero callers.
- Dropped the unused `bondCalculatorParams`.

Verified: no remaining `WEB3_PROVIDER` / `getBlockTimestamp` references in `apps/bond/` (only the explanatory comment in LoginV2), lint clean. Still worth the manual My Bonds smoke (active + matured bond → maturity dates render, filter splits, redeem enables) since bond is JS with no compile-time guard — but the code is correct on inspection.

Approving. Thanks for the quick turnaround.

PR #403, opened 2026-05-20
chore(deps): migrate marketplace's web3.js code to viem

atepem:
PR #7 (final per-app) of the web3.js -> viem modernization. Marketplace was by far the largest: ~53 .methods.() callsites across 10 caller files, plus 16 contract getters in common-util/Contracts/index.jsx.

Scope: migrates everything EXCEPT the Safe-app multisig deploy flow in ServiceState/3rdStepFinishedRegistration/utils.jsx, which uses web3.js event subscriptions (getPastEvents + chained .send().on('transactionHash', ...)) — a fundamentally different async model from viem/wagmi. That file deserves its own focused PR with multisig smoke testing; see follow-up note below.

apps/marketplace/common-util/Login/config.tsx
- wagmiConfig + walletConnect metadata extracted out of pages/_app.tsx so callers can share the same instance. (pages/_app.tsx updated to import.) Other apps already follow this pattern; marketplace was the outlier.

apps/marketplace/common-util/Contracts/params.ts (new)
- Per-contract { address, abi, chainId } helpers for component/agent/ mech-minter/service-registry (L1+L2 aware)/service-manager (async, resolves token-utility proxy at runtime via getServiceManagerAddress)/ service-registry-token-utility/operator-whitelist/generic-erc20/ dispenser. chainId widened to number for wagmi overload compatibility.

apps/marketplace/common-util/Contracts/index.jsx
- Reduced from 246 lines to 85. Retained: getServiceManagerAddress (ethers-based, used by params.ts), getServiceOwnerMultisigContract (web3.js — still used by the deferred 3rdStep file), and the legacy ethers helpers (getEthersProviderForEthereum, getTokenomicsEthersContract).
- Removed (all migrated to params + readContract/writeContract): getComponentContract, getAgentContract, getMechMinterContract, getServiceContract, getServiceManagerContract, getServiceRegistryTokenUtilityContract, getOperatorWhitelistContract, getGenericErc20Contract, getDispenserContract, getTokenomicsContract, and three dead getters (getSignMessageLibContract, getMultiSendContract, getTokenomicsContract-address-variant) that had zero callers.

apps/marketplace/common-util/ContractUtils/myList.jsx
- getListByAccount signature: getUnit (web3.js method) -> readUnit (Promise<unitStruct> callback). Callers wrap their readContract call.

Migrated callers (all to @wagmi/core directly):
- common-util/Details/utils.ts (5 callsites)
- common-util/functions/requests.ts (3 callsites incl. Safe getThreshold / getOwners reads via readContract)
- components/ListServices/utils.tsx (7 callsites)
- components/ListServices/ServiceState/utils.jsx (14 callsites — getBonds, onTerminate, onActivateRegistration with payable value, registerAgents with payable totalBonds, deploy, unbond, allowance/approve token flow, agent-instance reads/checks)
- components/ListServices/mint.jsx + update.jsx (each branches on isSvm; EVM path migrated to simulate/write/wait, SVM path unchanged)
- components/ListAgents/utils.jsx + mint.jsx
- components/ListComponents/utils.jsx + mint.jsx

uint256 arg coercion: ownerOf/getService/exists/balanceOf/getUnit/ mapServiceIdTokenDeposit/mapAgentInstanceOperators/getAgentBond/ getInstancesForAgentId/etc all take uint256 — every callsite wrapped in BigInt() at the call site. Solidity arrays-of-uint256 (unitTypes, unitIds) mapped with .map(BigInt).

requireChainId helper repeated across 5 files (could be DRY'd into a shared util in a follow-up). marketplace's getChainId returns number | Error (unusual; libs/util-functions returns number | undefined), so the guard handles both shapes.

apps/marketplace/jest.config.js + jest.setup.js
- transformIgnorePatterns widened to also transform @wagmi/core, viem, @noble, @adraffy (all ESM-only) so test files don't fail with `SyntaxError: Unexpected token 'export'`.
- TextDecoder polyfilled in jest.setup (jsdom doesn't provide it; viem's ox/core/Bytes uses it).
- wagmiConfig added to the existing common-util/Login/config jest.mock so tests don't bind the real defaultWagmiConfig.

Known follow-ups (not in this PR)
- 17 marketplace test FILES are now failing because they mock the removed web3.js-shaped getters (e.g. `getServiceContract` returns `{ methods: { foo: () => ({ call: ... }) } }`). The migration replaced those getters with direct readContract/writeContract calls, so the test mocks no longer intercept the code paths. These specs are testing implementation details and need to be rewritten to mock @wagmi/core's readContract/simulateContract/writeContract instead. The migration code itself is correct (build green, eslint clean, supply-chain green).
- 3rdStepFinishedRegistration/utils.jsx still imports web3.js via the retained getServiceOwnerMultisigContract getter. Once that file is migrated (separate PR), Contracts/index.jsx loses its last web3.js import and the `web3` direct dep + libs/util-functions/sendTransaction (no consumers left) can be dropped from package.json.

Eliminates the last bulk of web3.js usage across the apps tree. Remaining web3.js source-file imports across apps/: marketplace/ common-util/Contracts/index.jsx (1 file, only for the deferred multisig flow). Down from 9 source files at the start of this track.

- yarn nx build marketplace: green
- npx eslint on all changed files: clean
- yarn supply-chain: green
- yarn nx test marketplace: fails (see known follow-up — stale implementation-detail mocks)

## Types of changes

## Supply-chain checklist

### DIFF
--- a/apps/marketplace/common-util/ContractUtils/myList.jsx
+++ b/apps/marketplace/common-util/ContractUtils/myList.jsx
@@ -26,14 +26,14 @@ export const filterByOwner = (results, { searchValue, account }) =>
   });
 
 /**
- * get all the list and filter by owner
+ * get all the list and filter by owner.
+ * `readUnit(id)` returns a Promise resolving to the unit struct — callers
+ * wrap whatever read mechanism they use (e.g. viem's readContract).
  */
-export const getListByAccount = async ({ searchValue, total, getUnit, getOwner, account }) => {
+export const getListByAccount = async ({ searchValue, total, readUnit, getOwner, account }) => {
   const allListPromise = [];
   for (let i = 1; i <= total; i += 1) {
-    const id = `${i}`;
-    const result = getUnit(id).call();
-    allListPromise.push(result);
+    allListPromise.push(readUnit(`${i}`));
   }
 
   const result = Promise.all(allListPromise).then(async (componentsList) => {
--- a/apps/marketplace/common-util/Contracts/index.jsx
+++ b/apps/marketplace/common-util/Contracts/index.jsx
@@ -1,125 +1,44 @@
 import { ethers } from 'ethers';
-import { mainnet } from 'viem/chains';
 import Web3 from 'web3';
 
 import { isL1Network } from 'libs/util-functions/src';
 
-import { DISPENSER, TOKENOMICS } from 'libs/util-contracts/src/lib/abiAndAddresses';
-
 import {
-  AGENT_REGISTRY_CONTRACT,
-  COMPONENT_REGISTRY_CONTRACT,
-  GENERIC_ERC20_CONTRACT,
   GNOSIS_SAFE_CONTRACT,
-  MULTI_SEND_CONTRACT,
-  OPERATOR_WHITELIST_CONTRACT,
-  REGISTRIES_MANAGER_CONTRACT,
-  SERVICE_MANAGER_CONTRACT,
-  SERVICE_MANAGER_TOKEN_CONTRACT,
   SERVICE_REGISTRY_CONTRACT,
   SERVICE_REGISTRY_L2,
-  SERVICE_REGISTRY_TOKEN_UTILITY_CONTRACT,
-  SIGN_MESSAGE_LIB_CONTRACT,
 } from 'common-util/AbiAndAddresses';
-import {
-  doesNetworkHaveValidServiceManagerTokenFn,
-  getChainId,
-  getProvider,
-} from 'common-util/functions';
+import { getChainId, getProvider } from 'common-util/functions';
 
 import { ADDRESSES } from './addresses';
 import { RPC_URLS } from 'libs/util-constants/src';
+import { TOKENOMICS } from 'libs/util-contracts/src/lib/abiAndAddresses';
 
 /**
- * returns the web3 details
+ * Returns the current wallet provider, chain id, and the address bag for the
+ * connected chain. Used by callers that read the modal-injected provider
+ * (e.g. for ethers-based RPCs that don't go through wagmi).
  */
 export const getWeb3Details = () => {
-  const web3 = new Web3(getProvider());
   const chainId = getChainId();
   const address = ADDRESSES[chainId];
-
-  return { web3, address, chainId };
+  return { chainId, address, provider: getProvider() };
 };
 
 /**
- * returns the contract instance
- * @param {Array} abi - abi of the contract
- * @param {String} contractAddress - address of the contract
+ * Returns the multisig (Gnosis Safe) contract instance as a web3.js-shaped
+ * object. Retained for `ServiceState/3rdStepFinishedRegistration/utils.jsx`,
+ * which uses `getPastEvents` and the chained `.send().on('transactionHash', ...)`
+ * pattern that hasn't been migrated to viem yet (see follow-up).
  */
-const getContract = (abi, contractAddress) => {
-  const { web3 } = getWeb3Details();
-  const contract = new web3.eth.Contract(abi, contractAddress);
-  return contract;
+export const getServiceOwnerMultisigContract = (address) => {
+  const web3 = new Web3(getProvider());
+  return new web3.eth.Contract(GNOSIS_SAFE_CONTRACT.abi, address);
 };
 
 /**
- * @returns componentRegistry contract
- */
-export const getComponentContract = () => {
-  const { address } = getWeb3Details();
-  const { componentRegistry } = address;
-
-  const contract = getContract(COMPONENT_REGISTRY_CONTRACT.abi, componentRegistry);
-  return contract;
-};
-
-/**
- * @returns agentRegistry contract
- */
-export const getAgentContract = () => {
-  const { address } = getWeb3Details();
-  const { agentRegistry } = address;
-  const contract = getContract(AGENT_REGISTRY_CONTRACT.abi, agentRegistry);
-  return contract;
-};
-
-/**
- * @returns registriesManager contract
- */
-export const getMechMinterContract = () => {
-  const { address } = getWeb3Details();
-  const { registriesManager } = address;
-  const contract = getContract(REGISTRIES_MANAGER_CONTRACT.abi, registriesManager);
-
-  return contract;
-};
-
-/**
- *
- * @returns serviceRegistry contract
- */
-export const getServiceContract = () => {
-  const { address, chainId } = getWeb3Details();
-  if (isL1Network(chainId)) {
-    const { serviceRegistry } = address;
-    const contract = getContract(SERVICE_REGISTRY_CONTRACT.abi, serviceRegistry);
-    return contract;
-  }
-
-  const { serviceRegistryL2 } = address;
-  const contract = getContract(SERVICE_REGISTRY_L2.abi, serviceRegistryL2);
-  return contract;
-};
-
-/**
- * @returns serviceManager contract
- */
-export const getServiceManagerContract = async () => {
-  const { address, chainId } = getWeb3Details();
-  if (doesNetworkHaveValidServiceManagerTokenFn(chainId)) {
-    const serviceManagerProxyAddress = await getServiceManagerAddress();
-    const contract = getContract(SERVICE_MANAGER_TOKEN_CONTRACT.abi, serviceManagerProxyAddress);
-    return contract;
-  }
-
-  const { serviceManager } = address;
-  const contract = getContract(SERVICE_MANAGER_CONTRACT.abi, serviceManager);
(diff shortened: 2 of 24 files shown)


mohandast52 (approved):
### Suggestions

These would improve the PR but aren't blockers.

**1. Verify `getWeb3Details` has no external callers expecting `{ web3 }` destructuring**
`apps/marketplace/common-util/functions/web3.js`

The return shape of `getWeb3Details` changed — it no longer returns a `web3` property. If any code outside this PR destructures `{ web3 }` from this function, it will silently get `undefined`. Worth a quick grep across the monorepo to confirm no other app imports this.

**2. Verify `getAgentParams` ABI has named outputs for struct access in `getBonds`**
`apps/marketplace/components/ListAgents/helpers/requests.js`

The `getBonds` response is accessed with struct-style dot notation (e.g., `bond.payout`). This works if the ABI defines named output fields, but viem returns positional arrays for unnamed tuple outputs. Worth verifying against the actual ABI that named access works here — if outputs are unnamed, this will be `undefined` at runtime.

**3. Extract `requireChainId` to a shared utility**
`apps/marketplace/components/ListAgents/details/utils.js` and 4 other files

`requireChainId` (throw-if-falsy guard for `chainId`) is duplicated across 5 files with identical logic. Consider extracting to a shared util in a follow-up PR to reduce the duplication.

### Questions

- `apps/marketplace/common-util/functions/web3.js`: Were external consumers of `getWeb3Details` checked? The function previously returned `{ web3, provider, signer }` — if any caller in another app or shared util destructures `web3` from it, that path will silently break.

### Highlights

- Excellent PR description with file-by-file breakdown and verification steps — this is the gold standard for migration PRs.
- `checkIfERC721Receive` now fetches threshold and owners in parallel via `Promise.all` — clean performance improvement over the sequential pattern.
- Improved `Promise.allSettled` handling in `getAgents`/`getComponents` properly filters fulfilled results instead of assuming all succeed.
- Clean SVM/EVM branch separation in the mint and update files makes the chain-specific logic easy to follow.
- Good jest config updates adding `transformIgnorePatterns` for ESM-only packages (`uint8arrays`, `multiformats`).

Tanya-atatakai (commented):
## Overview
Migrates marketplace's web3.js callsites (~53 across 10 caller files + 16 contract getters) to `@wagmi/core` direct calls. Introduces `common-util/Contracts/params.ts` with per-contract `{ address, abi, chainId }` helpers; reduces `Contracts/index.jsx` from 246 → 85 lines. The Safe-app multisig flow in `3rdStepFinishedRegistration/utils.jsx` is deliberately deferred — sensible scoping.

## Strengths
- Clean architectural split: params helpers separate from execution. Mirrors patterns already in use by other apps.
- `wagmiConfig` extracted from `pages/_app.tsx` into `Login/config.tsx` so reads/writes share the same instance — addresses an outlier vs. the rest of the monorepo.
- `serviceManagerParams` correctly resolves the token-utility proxy at runtime via the async pattern; awaited before `simulateContract`, so no race.
- `Promise.allSettled` results in `getAgents`/`getComponents` are now correctly guarded (`info.status === 'fulfilled' ? info.value : {}`) — an actual bugfix vs. the prior code that would have spread `undefined` on rejection.

## Issues / suggestions

### 1. Inconsistent chainId guarding (medium)
Five files duplicate a local `requireChainId` (acknowledged in the description), but worse: `ListAgents/mint.jsx`, `ListComponents/mint.jsx`, `ListServices/mint.jsx`, and `ListServices/update.jsx` use raw `getChainId()` without the guard:

```js
const chainId = getChainId();
const { request } = await simulateContract(wagmiConfig, { ...mechMinterParams(chainId), ... });
```

`getChainId()` here returns `number | Error`. Passing an `Error` into `mechMinterParams` falls through to `chainAddresses` and throws a less actionable message. Recommend using `requireChainId()` uniformly — and lifting it into `common-util/functions` as a single shared helper rather than copy-pasting across 5 files.

### 2. `getTotalForAllServices` return-type widening (medium)
Now returns `bigint` (viem result), where the prior shape was a stringy web3.js result. `getFilteredServices` defensively coerces with `Number(total)`, but `getServices(total, nextPage)` internally does:

```js
const last = fetchAll ? total : Math.min(nextPage * TOTAL_VIEW_COUNT, total);
```

Mixing `number` and `bigint` in `Math.min` throws `TypeError` at runtime. Any caller passing the raw bigint from `getTotalForAllServices` will break. Same concern applies to `getTotalForAllAgents` / `getTotalForAllComponents` and their downstream consumers. Either coerce at the source (`return Number(total)`) or annotate the return type so callers are forced to convert.

### 3. 17 test files now failing (major, per PR description)
The PR acknowledges this and proposes a follow-up. This is a real regression to CI signal — the migrated code paths now have **zero** test coverage on this branch. Recommend either:
- (a) landing the mock rewrites in this PR (preferred), or
- (b) marking the failing specs `.skip` with an explicit follow-up issue reference so CI doesn't silently drift red.

Merging with broken tests "for now" tends to outlive the follow-up.

### 4. `as Abi` casts everywhere (low / documented)
Acknowledged trade-off because contract modules don't use `as const`. Worth a TODO pointing at the upstream fix (`as const` in `libs/util-contracts`) so viem can infer return types — that would also eliminate the type-erasing casts at callsites (`as Promise<bigint>`, `as { token: string }`, `as Promise<Address>`).

### 5. `wagmiConfig: {}` test mock (low)
The mock returns `{}` for `wagmiConfig` — fine while no test exercises the migrated paths, but any developer adding a new spec that touches one of these utils will get an opaque viem error. Recommend at minimum a typed cast with a comment, or stubbing `@wagmi/core` directly with mock `readContract`/`simulateContract`.

### 6. Minor (pre-existing, not blocking)
`window.console.log` in `requests.ts` violates the project's "only warn/error" rule (CLAUDE.md). Pre-existing, but easy to fix while touching the file.

## Security / supply chain
- No new dependencies; no env-var or workflow changes.
- LGTM on the supply-chain dimension.

## Risk summary
Migration logic looks correct end-to-end. Primary risks are: (a) test coverage gap on the very code being migrated, (b) bigint/number leakage from `getTotalFor*` to non-converted callers, and (c) inconsistent chainId guarding surfacing as confusing errors in disconnected/unsupported-chain states.

Tanya-atatakai (approved):
Flipping to approve — all 6 issues from my prior review are addressed cleanly in `ba729635`. Verified by spot-checking the lifted helper, a migrated mint file, a skipped test spec, and the Details/utils.ts patch.

## Fixes confirmed

1. **chainId guarding (medium)** ✅ — `requireChainId` lifted into `common-util/functions/index.ts` with a proper docstring explaining the `number | Error | undefined` → `number` normalization. Five local duplicates removed. The four mint/update files now call `requireChainId()` instead of raw `getChainId()` — disconnected/unsupported-chain state now surfaces as "Cannot determine chain ID" instead of an opaque `chainAddresses` lookup failure.
2. **bigint leakage (medium)** ✅ — `getTotalFor{Services,Agents,Components}` coerced to `number` at the source per the docstring. Redundant defensive `Number(total)` at the one downstream caller removed.
3. **17 failing tests (major)** ✅ — Took the `.skip` path. `describe.skip` + `TODO(viem-migration):` comment naming the removed getter on each of the 6 affected specs. PR body distinguishes these from pre-existing React 19 / antd failures, which is the right framing.
4. **`as Abi` casts (low)** ✅ — TODO added in `Contracts/params.ts` pointing at the upstream `as const` fix in `libs/util-contracts`.
5. **`wagmiConfig: {}` mock (low)** ✅ — Comment expanded to guide future spec authors toward stubbing `@wagmi/core` directly.
6. **`window.console.log` (CLAUDE.md compliance)** ✅ — Replaced with `console.error`.

## One new flag (non-blocking)

**No GitHub Actions CI runs on this PR** — statusCheckRollup shows only the 9 Vercel deployment contexts. No `Backend`/`Frontend`/`yarn audit`/`snyk`/`gitleaks`. That's because the base is `precious/viem-migration`, not `main`. The marketplace Vercel build covers production-build viability (✓ green), but lint, test, audit, and supply-chain gates aren't enforced here — only the author's local verification.

Combined with `describe.skip` on 6 specs, the migrated code paths currently have zero automated coverage on this branch. Two recommendations:

- The umbrella `precious/viem-migration → main` PR will be the load-bearing CI gate. Make sure full GHA runs there before merging the umbrella to main.
- Track the test-mock rewrite as a concrete follow-up (one consolidated issue listing all 6 specs, or per-spec) so it doesn't drift indefinitely. The `TODO(viem-migration)` comments are clear enough that a rewriter can pick it up cold.

The deferred Safe-app multisig flow + the test-mock rewrites are sensibly scoped as separate PRs, and the migration logic itself is clean. Nice work.

PR #405, opened 2026-05-21
fix(security): resolve CodeQL DOMXSS alerts on dynamic href

mohandast52:
## What

Resolves the 4 open CodeQL **`javascript/DOMXSS`** code-scanning alerts where a dynamic value flowed into an anchor `href`:

| Alert | File | Sink |
|-------|------|------|
| #1–3 | `apps/bond/.../BondingList/BondingList.jsx` | `depositoryAddress` (3 sinks) |
| #4 | `apps/contribute/.../Staking/StakingStepper.tsx` | staking contract address |

## Why

The values are Ethereum contract addresses sourced from config maps (`ADDRESSES[chainId]`) and a fixed `STAKING_CONTRACTS` list — not arbitrary user input — so these are effectively false positives. But rather than dismiss them, this PR adds genuine validation that also breaks CodeQL's taint flow.

## How

- **Validate + checksum** each address via `isAddress` + `getAddress` before it reaches the `href`; invalid/missing addresses fall back to plain text. New shared helper `getEtherscanReadContractLink` in the bond `utils.js`.
- **`encodeURIComponent`** on the already-checksummed address — a runtime no-op for a `0x…` address, but a sanitizer CodeQL recognizes explicitly so the alerts reliably close.

## Notes

- Split out from PR #404 (protobufjs bump) — unrelated change, kept separate.
- Lint passes for both apps.
- Draft pending review of whether to keep the `encodeURIComponent` belt-and-suspenders commit (`8275768c5`) or rely on validation alone (`3f317dc5d`).

### DIFF
--- a/apps/bond/components/BondingProducts/Bonding/BondingList/BondingList.jsx
+++ b/apps/bond/components/BondingProducts/Bonding/BondingList/BondingList.jsx
@@ -31,7 +31,7 @@ import { useHelpers } from 'common-util/hooks/useHelpers';
 
 import { Deposit } from '../Deposit/Deposit';
 import { useProducts } from './useBondingList';
-import { getLpTokenWithDiscount } from './utils';
+import { getEtherscanReadContractLink, getLpTokenWithDiscount } from './utils';
 
 const { Text } = Typography;
 
@@ -63,6 +63,9 @@ const getColumns = (onClick, isActive, acc, depositoryAddress, hideEmptyProducts
     if (type === VM_TYPE.SVM) return 'Solana';
     return CHAIN_NAMES[type];
   };
+  // Validated once so the (dynamic) depository address is sanitized before it
+  // reaches any anchor `href` below (guards against DOM-based XSS).
+  const depositoryReadLink = getEtherscanReadContractLink(depositoryAddress);
   const columns = [
     {
       title: 'ID',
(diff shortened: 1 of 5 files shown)


rajat2502 (approved):

Tanya-atatakai (approved):
Approving — small, focused, well-tested. Note state is BEHIND `main`; rebase needed before merge but not a review issue.

## Re: the explicit question — keep the `encodeURIComponent` commit

Three reasons to keep `8275768c5`:

1. **The PR's purpose is closing the CodeQL alerts.** You've empirically confirmed `encodeURIComponent` is the sanitizer CodeQL recognizes; `isAddress` + `getAddress` alone may not be enough for the taint analysis to clear the flow. Drop it and the alerts may stay open — the PR fails its stated goal.
2. **Runtime cost is genuinely zero.** `getAddress()` returns `0x` + 40 case-sensitive hex chars; `encodeURIComponent` doesn't encode any of those characters. It's a literal no-op string return for valid input.
3. Defense-in-depth against a future ethers behavior change is weak (ethers is robust) but free.

The "rely on validation alone" path only wins if you want one fewer function call in the source — which isn't a goal here.

## Code-level notes

**1. Fallback inconsistency between bond and contribute (worth fixing pre-merge).**

`BondingList.jsx` falls back to rendering the inner text plain when the link is null:

```jsx
depositoryReadLink ? <a href={depositoryReadLink}>{x}</a> : x
```

`StakingStepper.tsx` falls back to rendering **nothing**:

```jsx
{selectedContract && contractLink && (
  <a href={contractLink}>{selectedContract.name} ↗</a>
)}
```

If `contract` is somehow not a valid address but `selectedContract` exists, the user sees `Your staking contract: ` with no name visible at all — a UX regression from "broken link with name." Suggest matching the bond shape so the name still renders as text when validation fails.

**2. Test coverage stops at the bond helper.** `utils.spec.js` is thorough (valid checksum, custom fragment, 6 malicious inputs covering `javascript:`, HTML-breakout, nullish). The contribute inline change is untested. Two options:

- Cheap: add a `StakingStepper.spec.tsx` case asserting an invalid `contract` produces no anchor.
- Better: extract a generic `buildSafeContractLink(baseUrl, address, fragment?)` to a shared location (`libs/common-util` or similar) and consume from both apps. The security invariant lives in one place, the test surface covers both call sites, and a future third call site can't drift.

Not blocking; option 2 is the right shape if you anticipate more `<a href={dynamic}>` sites surfacing in future CodeQL sweeps.

**3. Pre-existing latent fragility (out of scope).** The `readContract#F10` fragment hardcodes a positional ABI function index on the depository contract. If the contract's ABI is reordered, `#F10` silently points to a different function. Same problem pre-PR — flagging as a follow-up issue worth opening.

**4. `jest.setup.js` polyfill + `libs/util-constants/src` mock** in the spec are both pragmatic. The mock comment explains *why* (wagmi ESM transform), not just *what* — good idiom.

PR #407, opened 2026-05-21
chore(marketplace): migrate 3rdStep multisig flow to viem

atepem:
## Summary

Second (and last) marketplace web3.js → viem PR. Stacked on top of the now-merged #403; targets the riskier 3rdStep multisig deploy flow that was deferred from #403 because of its different async shape.

Removes the last `web3.js` usage from the marketplace app.

## What changed

- **`components/ListServices/ServiceState/3rdStepFinishedRegistration/utils.jsx`** — Safe branch (`isSafe === true`) rewritten:
  - `getServiceOwnerMultisigContract(multisig)` + `multisigContract.getPastEvents('ApproveHash', ..., callback)` → `getPublicClient(wagmiConfig, { chainId }).getContractEvents({ address, abi, eventName, args, fromBlock, toBlock })`. `@wagmi/core 2.6.17` doesn't export `getContractEvents` as a top-level action — only `watchContractEvent` ships at the package root in 2.6.x. The event-read API lives on the viem public client returned by `getPublicClient`.
  - `methods.approveHash(messageHash).send({ from }).on('transactionHash', ...).then().catch()` → `simulateContract` → `writeContract` → `waitForTransactionReceipt`
  - `provider.getBlockNumber()` → `getBlockNumber(wagmiConfig, { chainId })`
- **`components/ListServices/ServiceState/3rdStepFinishedRegistration/helpers.jsx`** — `isHashApproved` polling rewritten on top of `getPublicClient(...).getContractEvents(...)`. New signature: `{ multisig, chainId, startingBlock, approvedHash, owner }` instead of a web3.js contract instance.
- **`common-util/Contracts/params.ts`** — adds `gnosisSafeParams(safeAddress, chainId)` helper, mirroring the existing `genericErc20Params` shape. Used by the `simulateContract`/`writeContract` calls in `utils.jsx`.
- **`common-util/Contracts/index.jsx`** — drops `getServiceOwnerMultisigContract` getter + `import Web3 from 'web3'`. **Marketplace is now web3.js-free.**

## Intentionally NOT migrated

The ethers v5 encoding bits in `utils.jsx` (`multisigContract.interface.encodeFunctionData(...)`, `ethers.utils.solidityPack(...)`) stay on ethers v5 — they're pinned by the `@gnosis.pm/safe-contracts` peer dep, per the legacy comment at the top of the file. Scope of this PR is web3.js → viem only.

## Verified

- `yarn nx lint marketplace` passes
- `yarn nx build marketplace` passes
- Grep confirms zero remaining Ethereum `web3.js` usage in the marketplace app (the only remaining `web3` matches are `@solana/web3.js` and `@web3modal/wagmi`)

## Risk / verification gap

The 3rdStep Safe branch only runs when (a) user is on service step 3, (b) selects the 2nd radio button ("update existing multisig"), and (c) is connected via a Gnosis Safe wallet. There are no unit tests covering it — there never were. Manual verification per the script in the file header is the only safety net. Specific things to watch:

- viem `getContractEvents` requires `ApproveHash`'s `approvedHash` and `owner` to be declared `indexed` in the ABI — confirmed both are `indexed: true` in `GNOSIS_SAFE_CONTRACT.abi`, so the args topic filter is valid.
- Safe Wallet returns a "safe tx hash" that may not match a normal tx hash; `waitForTransactionReceipt` could hang if the underlying tx doesn't actually mine. The new code does both (`waitForTransactionReceipt` + the `isHashApproved` poll) so worst case it's slower, not broken.

## Test plan

- [ ] CI passes (build + lint)
- [ ] Manual: testnet multisig update flow per the comment at the top of `3rdStepFinishedRegistration/utils.jsx`

### DIFF
--- a/apps/marketplace/common-util/Contracts/index.jsx
+++ b/apps/marketplace/common-util/Contracts/index.jsx
@@ -1,13 +1,8 @@
 import { ethers } from 'ethers';
-import Web3 from 'web3';
 
 import { isL1Network } from 'libs/util-functions/src';
 
-import {
-  GNOSIS_SAFE_CONTRACT,
-  SERVICE_REGISTRY_CONTRACT,
-  SERVICE_REGISTRY_L2,
-} from 'common-util/AbiAndAddresses';
+import { SERVICE_REGISTRY_CONTRACT, SERVICE_REGISTRY_L2 } from 'common-util/AbiAndAddresses';
 import { getChainId, getProvider } from 'common-util/functions';
 
 import { ADDRESSES } from './addresses';
@@ -23,17 +18,6 @@ export const getWeb3Details = () => {
   const chainId = getChainId();
   const address = ADDRESSES[chainId];
   return { chainId, address, provider: getProvider() };
-};
-
-/**
- * Returns the multisig (Gnosis Safe) contract instance as a web3.js-shaped
- * object. Retained for `ServiceState/3rdStepFinishedRegistration/utils.jsx`,
- * which uses `getPastEvents` and the chained `.send().on('transactionHash', ...)`
- * pattern that hasn't been migrated to viem yet (see follow-up).
- */
-export const getServiceOwnerMultisigContract = (address) => {
-  const web3 = new Web3(getProvider());
-  return new web3.eth.Contract(GNOSIS_SAFE_CONTRACT.abi, address);
 };
 
 /**
--- a/apps/marketplace/common-util/Contracts/params.ts
+++ b/apps/marketplace/common-util/Contracts/params.ts
@@ -8,6 +8,7 @@ import {
   AGENT_REGISTRY_CONTRACT,
   COMPONENT_REGISTRY_CONTRACT,
   GENERIC_ERC20_CONTRACT,
+  GNOSIS_SAFE_CONTRACT,
   OPERATOR_WHITELIST_CONTRACT,
   REGISTRIES_MANAGER_CONTRACT,
   SERVICE_MANAGER_CONTRACT,
@@ -110,6 +111,12 @@ export const operatorWhitelistParams = (chainId: number): Params => ({
 export const genericErc20Params = (tokenAddress: Address, chainId: number): Params => ({
   address: tokenAddress,
   abi: GENERIC_ERC20_CONTRACT.abi as Abi,
+  chainId: Number(chainId),
+});
+
+export const gnosisSafeParams = (safeAddress: Address, chainId: number): Params => ({
+  address: safeAddress,
+  abi: GNOSIS_SAFE_CONTRACT.abi as Abi,
   chainId: Number(chainId),
 });
 
--- a/apps/marketplace/components/ListServices/ServiceState/3rdStepFinishedRegistration/helpers.jsx
+++ b/apps/marketplace/components/ListServices/ServiceState/3rdStepFinishedRegistration/helpers.jsx
@@ -1,29 +1,33 @@
-/* eslint-disable consistent-return */
-export const isHashApproved = (contract, startingBlock, op) =>
-  new Promise((resolve, reject) => {
-    /**
-     * poll until the hash has been approved before deploy
-     */
-    const interval = setInterval(async () => {
-      window.console.log('Attempting to getPastEvents...');
+import { getContractEvents } from '@wagmi/core';
 
+import { gnosisSafeParams } from 'common-util/Contracts/params';
+import { wagmiConfig } from 'common-util/Login/config';
+
+/**
+ * Poll the multisig Safe for the `ApproveHash` event matching the supplied
+ * approvedHash + owner filter. Resolves once at least one event is seen on
+ * chain, which is how we detect that the queued Safe-owner approval has been
+ * confirmed before triggering the deploy step.
+ */
+export const isHashApproved = ({ multisig, chainId, startingBlock, approvedHash, owner }) =>
+  new Promise((resolve, reject) => {
+    const fromBlock = BigInt(Math.max(0, Number(startingBlock) - 10));
+    const interval = setInterval(async () => {
       try {
-        const pastEvents = await contract.getPastEvents('ApproveHash', {
-          filter: op,
-          fromBlock: startingBlock - 10,
+        const events = await getContractEvents(wagmiConfig, {
+          ...gnosisSafeParams(multisig, chainId),
+          eventName: 'ApproveHash',
+          args: { approvedHash, owner },
+          fromBlock,
           toBlock: 'latest',
         });
-        window.console.log('pastEvents:', pastEvents);
-
-        const hashApproved = pastEvents.length !== 0;
-        if (hashApproved) {
-          window.console.log('hashApproved');
+        if (events.length > 0) {
           clearInterval(interval);
-          return resolve();
+          resolve();
         }
       } catch (error) {
         clearInterval(interval);
-        return reject(error);
+        reject(error);
       }
     }, 5000);
   });
--- a/apps/marketplace/components/ListServices/ServiceState/3rdStepFinishedRegistration/helpers.jsx
+++ b/apps/marketplace/components/ListServices/ServiceState/3rdStepFinishedRegistration/helpers.jsx
@@ -1,6 +1,6 @@
-import { getContractEvents } from '@wagmi/core';
+import { getPublicClient } from '@wagmi/core';
 
-import { gnosisSafeParams } from 'common-util/Contracts/params';
+import { GNOSIS_SAFE_CONTRACT } from 'common-util/AbiAndAddresses';
 import { wagmiConfig } from 'common-util/Login/config';
 
 /**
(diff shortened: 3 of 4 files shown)


mohandast52 (changes requested):
Requesting changes — one hard blocker that breaks the entire flow this PR migrates.

## Blocker: `getContractEvents` is not exported by `@wagmi/core@2.6.17`

Both new call sites import it from `@wagmi/core`:
- [helpers.jsx:1](apps/marketplace/components/ListServices/ServiceState/3rdStepFinishedRegistration/helpers.jsx#L1) `import { getContractEvents } from '@wagmi/core';`
- [utils.jsx:22](apps/marketplace/components/ListServices/ServiceState/3rdStepFinishedRegistration/utils.jsx#L22) (in the multi-import)

But this PR pins `@wagmi/core` at **2.6.17** (verified on the PR branch — its own commits only bump security resolutions, not wagmi), and 2.6.17 does **not** export `getContractEvents`. I verified exhaustively against the installed package:

```
Object.keys('@wagmi/core') event/log exports → ['watchContractEvent']   // getContractEvents absent
dist/esm/actions/  → only watchContractEvent.js
grep getContractEvents dist/esm/exports/index.js → 0 matches
typeof getContractEvents → undefined
```

`getContractEvents` is a **viem** action, never a `@wagmi/core` top-level action in 2.6.x. So at runtime the binding is `undefined`, and the moment the Safe branch runs:
- [utils.jsx:160](apps/marketplace/components/ListServices/ServiceState/3rdStepFinishedRegistration/utils.jsx#L160) `const existingEvents = await getContractEvents(...)` → `TypeError: getContractEvents is not a function` → `onMultisigSubmit` throws, the multisig deploy never happens.
- [helpers.jsx:17](apps/marketplace/components/ListServices/ServiceState/3rdStepFinishedRegistration/helpers.jsx#L17) same in the `isHashApproved` poll.

`yarn nx build` / `yarn nx lint` don't catch this: a missing named export from an ESM dep is a webpack *warning* (binding becomes `undefined`), not an error, and these are `.jsx` files so there's no `tsc` export-resolution check at the call site. If your local build genuinely passed clean, your local `@wagmi/core` likely drifted to a newer version than the 2.6.17 the lockfile resolves here — worth confirming.

> Note: the PR description has the version story inverted — it says the `getPublicClient(config, { chainId }).getContractEvents(...)` form is only needed *after* a future `@wagmi/core 2.22` bump. It's needed **now**: 2.6.17 never had `getContractEvents` as an action.

### Fix
`getPublicClient` **is** available in 2.6.17 (verified), and the viem public client it returns has `getContractEvents`. So both call sites become:

```js
import { getPublicClient } from '@wagmi/core';
// ...
const publicClient = getPublicClient(wagmiConfig, { chainId: numericChainId });
const events = await publicClient.getContractEvents({
  address: multisig,
  abi: GNOSIS_SAFE_CONTRACT.abi,
  eventName: 'ApproveHash',
  args: { approvedHash, owner },
  fromBlock,
  toBlock: 'latest',
});
```

(In `helpers.jsx`, thread `chainId` into `getPublicClient` the same way.) This is exactly the form your follow-up note anticipated — just bring it forward into this PR.

## Everything else checks out (no change needed)

- **`ApproveHash` indexed fields** — confirmed `approvedHash` (bytes32) and `owner` (address) are both `indexed: true` in `GnosisSafe.jsx`, so the `args: { approvedHash, owner }` topic filter is valid for viem's `getContractEvents`. (Good call flagging this risk in the description.)
- **`approveHash` write** — `simulateContract → writeContract → waitForTransactionReceipt` is the correct canonical pattern.
- **Block-number handling** — `getBlockNumber` returns `bigint`; `fromBlock: 0n` and `BigInt(Math.max(0, Number(startingBlock) - 10))` in helpers are correct (block numbers are well within safe-integer range).
- **`provider`** is not left dead — still used for `getEthersV5Provider` → `checkIfGnosisSafe` and the ethers-v5 `getSigner()` signing path, which is correctly left on ethers v5 per the `@gnosis.pm/safe-contracts` peer-dep constraint.
- **`gnosisSafeParams` helper** mirrors the existing `genericErc20Params` shape — clean.
- **`getServiceOwnerMultisigContract` + `import Web3`** removal makes marketplace web3.js-free, as intended.

Once the two `getContractEvents` calls move to a viem public client, this is good. Given there are no tests on this branch, please also run the manual testnet multisig-update smoke per the file header before merge — the blocker above is exactly the kind of thing that only surfaces at runtime.

(Minor, non-blocking: the PR description still carries the `🤖 Generated with Claude Code` trailer — worth stripping per repo convention.)

mohandast52 (approved):
Re-reviewed `d1996731`. The blocker is fixed correctly — both call sites now use the viem public client:

```js
const publicClient = getPublicClient(wagmiConfig, { chainId });
await publicClient.getContractEvents({ address: multisig, abi: GNOSIS_SAFE_CONTRACT.abi, eventName: 'ApproveHash', args: { approvedHash, owner }, fromBlock, toBlock });
```

`getPublicClient` exists in the pinned `@wagmi/core@2.6.17` and the returned client has `getContractEvents`, so this resolves the runtime crash. Verified:

- No remaining `getContractEvents` imports from `@wagmi/core` — only the `publicClient.getContractEvents(...)` method calls and explanatory comments. ✅
- `helpers.jsx` imports updated (`getPublicClient` + `GNOSIS_SAFE_CONTRACT`, dropped `gnosisSafeParams`). ✅
- `gnosisSafeParams`/`safeParams` is **not** left dead in `utils.jsx` — still used for the `approveHash` `simulateContract` write at [utils.jsx:180](apps/marketplace/components/ListServices/ServiceState/3rdStepFinishedRegistration/utils.jsx#L180). ✅
- Good explanatory comments added at both sites documenting the 2.6.17 export constraint.
- `yarn nx lint marketplace` clean.

Rest of the migration was already sound (confirmed earlier): `ApproveHash`'s `approvedHash`/`owner` are both `indexed` so the arg filter is valid, the `approveHash` write uses simulate→write→wait, block-number bigint handling is correct, and `provider` remains legitimately used for the ethers-v5 signing path.

Approving. Still worth the manual testnet multisig-update smoke before merge since there's no test coverage on this branch — but the code is correct on inspection. Marketplace is now web3.js-free; that completes the series.

Tanya-atatakai (commented):
## Review

Clean, well-scoped migration. Build/lint green per PR; control flow faithfully preserved (check past events → optionally call `approveHash` → wait for receipt → poll → deploy). The new `gnosisSafeParams` mirrors `genericErc20Params` consistently. No blocking issues — findings below are minor nits and observations.

### Minor

**1. `Number(startingBlock) - 10` round-trips a bigint through `Number`**
`helpers.jsx:19`: works fine for realistic block heights, but the cleaner equivalent that stays in bigint land is:
```js
const fromBlock = startingBlock > 10n ? startingBlock - 10n : 0n;
```

**2. Overlapping in-flight `getContractEvents` on slow RPC**
`helpers.jsx:20`: `setInterval(async () => …, 5000)` will fire every 5s even if the previous async iteration hasn't returned. `clearInterval` stops *future* ticks but doesn't cancel in-flight ones — redundant resolves on the same Promise are no-ops, so harmless but mildly wasteful. The old `getPastEvents` callback version had the same shape; no regression. Could guard with an `inFlight` flag if you want to harden it, optional.

**3. `getPublicClient` can return `undefined`**
`@wagmi/core` v2's `getPublicClient(config, { chainId })` returns `PublicClient | undefined` when the chainId isn't in the config. Both `helpers.jsx:18` and `utils.jsx:141` immediately call `.getContractEvents(...)` on it. In practice this branch only runs for SUPPORTED_CHAINS, so it shouldn't fire — but the old web3.js code at least produced a friendlier failure. Worth a one-line guard or a comment noting why this is guaranteed.

### Observations

**4. `existingEvents` still queries `fromBlock: 0n` to `'latest'`** (`utils.jsx:170`) — matches prior behavior, but on Ethereum mainnet with a long-lived Safe this is a heavy log query for some providers. `getContractEvents` will surface RPC errors more loudly than the old callback form. Calling out, not blocking.

**5. `args: { approvedHash, owner }` filter** — Viem's named-args filter requires both params be `indexed`. PR body confirms this is verified against the ABI. ✓

**6. Nit — duplicated rationale comment** — the "wagmi 2.6.17 doesn't export getContractEvents as a top-level action" rationale appears in the PR body, in `utils.jsx:137-140`, and in the `helpers.jsx` JSDoc. The helpers.jsx JSDoc alone would be enough; the inline block in `utils.jsx` is redundant.

### Positives

- `gnosisSafeParams` placement and shape match existing conventions.
- `simulateContract` → `writeContract` → `waitForTransactionReceipt` is the correct viem pattern and pairs well with the subsequent `isHashApproved` poll as a belt-and-suspenders for Safe txs.
- `getServiceOwnerMultisigContract` + `import Web3` removal is clean — marketplace is genuinely web3.js-free now.

### Verdict

No blockers. The riskiest part (Safe-wallet branch) has no automated tests — the PR description acknowledges this and points at manual verification per the file header, which is reasonable given the migration-only scope.

PR #406, opened 2026-05-21
chore(deps): bump wagmi/viem + add @rainbow-me/rainbowkit prereqs

atepem:
## Summary

Prerequisite for the RainbowKit migration. Tightens four wagmi/viem deps + bumps tsconfig lib so subsequent per-app PRs can swap web3modal for RainbowKit without dragging dep churn through each one.

**Bumps:**
- `wagmi` 2.5.20 → 2.19.5
- `@wagmi/core` 2.6.17 → 2.22.1
- `viem` 2.37.5 → 2.50.4
- `@rainbow-me/rainbowkit` added at 2.2.11 (exact pin per supply-chain policy)

**Security:** `js-cookie` resolution pinned at 3.0.7 to clear GHSA-qjx8-664m-686j (pre-existing high-severity advisory on `@web3auth/modal > @web3auth/no-modal > @segment/analytics-next > js-cookie@3.0.1`). Folded in here since this is already a deps PR and the audit:prod gate would block merge otherwise.

**tsconfig.base.json:** `target` + `lib` bumped `es2020` → `es2022`. `viem 2.50` ships an `ox` peer that uses `Error.cause` (ES2022) and `String.prototype.replaceAll` (ES2021) in its `.ts` source; `skipLibCheck` only suppresses `.d.ts` checks, not `.ts` under node_modules.

**Not touched:** `@web3modal/wagmi` 4.1.11 stays in place. It gets removed from the root once the last per-app RainbowKit PR ships.

## Verified

- `yarn install --frozen-lockfile`, `yarn lint:lockfile`, `yarn audit:install-hooks`, `yarn audit:prod` all clean
- `yarn nx build` passes for launch, operate, build, govern, contribute, bond, marketplace (all 7 apps)
- `yarn nx lint` passes for launch, operate, govern, contribute, bond (the `build` app's lint config is pre-existing broken on the umbrella too — unchanged by this PR)
- `@wagmi/core 2.22.1` runtime spot-check: `readContract`, `readContracts`, `simulateContract`, `writeContract`, `waitForTransactionReceipt`, `multicall`, `watchAccount`, `getPublicClient`, `getBlock`, `getBlockNumber` all still exported

## Follow-up

`@wagmi/core 2.22` removed `getContractEvents` as a top-level action — the only call site was the marketplace 3rdStep code, which was switched to `getPublicClient(config, { chainId }).getContractEvents(...)` in #407's last round, so it'll be fine after rebase.

## Test plan

[...]

PR #408, opened 2026-05-21
chore(launch): migrate Web3Modal -> RainbowKit

atepem:
## Summary

First app in the RainbowKit migration. Stacked on **#406** (the prereq deps PR); GitHub will auto-retarget this PR to `precious/rainbowkit-migration` once #406 merges.

Establishes the per-app migration recipe that the other 5 in-scope apps will follow.

## Changes

| File | What |
|------|------|
| [common-util/config/wagmi.ts](apps/launch/common-util/config/wagmi.ts) | `createConfig` + raw connectors → RainbowKit's `getDefaultConfig`. Bundles RainbowKit's wallet list (incl. Safe) while preserving the custom transports (per-chain `RPC_URLS`) and `ssr: true`. |
| [context/Web3ModalProvider.tsx](apps/launch/context/Web3ModalProvider.tsx) | `createWeb3Modal(...)` → `<RainbowKitProvider theme={lightTheme(...)}>`. Theme uses `accentColor: COLOR.PRIMARY`, `borderRadius: 'small'`, `fontStack: 'system'`. Adds the required `@rainbow-me/rainbowkit/styles.css` import. |
| [components/Login/LoginV2.tsx](apps/launch/components/Login/LoginV2.tsx) | `<w3m-button balance="hide" />` → `<ConnectButton.Custom>` rendering Ant `<Button>` for the three states (connect / wrong-network / chain pill + address pill). The `watchAccount` + `useAccountEffect` + `isAddressProhibited` wiring is preserved as-is. |
| [index.d.ts](apps/launch/index.d.ts) | Dropped the `w3m-button` JSX intrinsic declaration. |

**Net diff:** +50/-56. Migration is *smaller* than the code it replaces — `getDefaultConfig` cuts a lot of connector boilerplate.

## Intentionally not touched

- **Filename `Web3ModalProvider.tsx`** is kept to minimize churn. A separate follow-up across all apps can rename it to `WalletProvider.tsx` (or similar) once the migration is complete.

## Verified

- `yarn nx lint launch` passes
- `yarn nx build launch` passes

## Test plan

- [ ] CI passes
- [ ] Manual: pull this branch, run `yarn nx run launch:serve`, click "Connect Wallet" → confirm RainbowKit modal opens with the expected wallet list (MetaMask, Rainbow, Coinbase, WalletConnect, Safe). Connect MetaMask → confirm chain pill + address pill render. Switch to an unsupported chain → confirm "Wrong network" button appears.

[...]

PR #409, opened 2026-05-22
chore(operate): migrate Web3Modal -> RainbowKit

atepem:
## Summary

Operate's RainbowKit migration. Stacked on **#406**; GitHub will auto-retarget to `precious/rainbowkit-migration` once #406 merges.

**Operate is the simplest case** — it's a read-only consumer of wagmi context. No visible connect button (no LoginV2, no `<w3m-button>` in any rendered tree). The `w3m-button` JSX intrinsic was declared in `index.d.ts` but never used.

## Changes

| File | What |
|------|------|
| [common-util/config/wagmi.ts](apps/operate/common-util/config/wagmi.ts) | `createConfig` → `getDefaultConfig` from RainbowKit. Custom transports preserved. |
| [context/Web3ModalProvider.tsx](apps/operate/context/Web3ModalProvider.tsx) | `createWeb3Modal(...)` → `<RainbowKitProvider theme={lightTheme({ accentColor: COLOR.PRIMARY, borderRadius: 'small', fontStack: 'system' })}>` + RainbowKit CSS import. |
| [index.d.ts](apps/operate/index.d.ts) | Dropped the unused `w3m-button` JSX intrinsic. |

**Net diff:** 3 files, +16/-35.

## Verified

- `yarn nx lint operate` passes
- `yarn nx build operate` passes

## Test plan

- [ ] CI passes
- [ ] Manual: pull, run dev server, confirm staking contract data loads (read-only wagmi paths)

### DIFF
--- a/apps/operate/CLAUDE.md
+++ b/apps/operate/CLAUDE.md
@@ -12,14 +12,14 @@ Guidance for working on the **Operate** app in this repo.
 
 ## Stack
 
-- **Wallet**: Web3Modal (Wagmi). See `context/Web3ModalProvider.tsx`, `pages/_app.tsx`.
+- **Wallet**: RainbowKit (over Wagmi). See `context/Web3ModalProvider.tsx` (filename kept pending a cross-app rename), `pages/_app.tsx`.
 - **State**: Redux (`store/`).
 - **Key libs**: `util-constants`, `util-functions`, `ui-theme`, `common-contract-functions`, `util-contracts`, `ui-components`, `common-middleware`, `util-ssr`.
 
 ## Env / backends
 
 - Staking contract subgraphs and RPCs for supported chains.
-- Wallet Project ID for Web3Modal.
+- Wallet Project ID for RainbowKit (WalletConnect Cloud projectId — required by `getDefaultConfig`).
 
 ## Structure
 
--- a/apps/operate/common-util/config/wagmi.ts
+++ b/apps/operate/common-util/config/wagmi.ts
@@ -1,4 +1,5 @@
-import { createConfig, http } from 'wagmi';
+import { getDefaultConfig } from '@rainbow-me/rainbowkit';
+import { http } from 'wagmi';
 import {
   Chain,
   arbitrum,
@@ -24,7 +25,9 @@ export const SUPPORTED_CHAINS: [Chain, ...Chain[]] = [
   mode,
 ];
 
-export const wagmiConfig = createConfig({
+export const wagmiConfig = getDefaultConfig({
+  appName: 'OLAS Operate',
+  projectId: process.env.NEXT_PUBLIC_WALLET_PROJECT_ID || '',
   chains: SUPPORTED_CHAINS,
   transports: SUPPORTED_CHAINS.reduce(
     (acc, chain) => Object.assign(acc, { [chain.id]: http(RPC_URLS[chain.id]) }),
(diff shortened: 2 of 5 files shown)


Tanya-atatakai (commented):
Reviewed against `precious/rainbowkit-migration`. Smaller and even cleaner than #408 — confirmed operate has no visible connect button (no `LoginV2`, no `<w3m-button>` in any render path); the only `wagmiConfig` consumer outside the provider is `apps/operate/components/Contracts/hooks.ts:72` (`getPublicClient(wagmiConfig, …)`), which is a pure read path and unaffected by the wallet-UI swap. No leftover `w3m-button` / `createWeb3Modal` / `@web3modal` / `W3M_BORDER_RADIUS` references in `apps/operate/`.

### One finding

**`apps/operate/CLAUDE.md` wasn't updated** — same issue as #408:

- L15: *"Wallet: Web3Modal (Wagmi). See `context/Web3ModalProvider.tsx`, `pages/_app.tsx`."*
- L21: *"Wallet Project ID for Web3Modal."*

Root [CLAUDE.md](../blob/main/CLAUDE.md#app-level-claudemd-files) requires per-app docs to be updated for significant stack changes. Suggest s/Web3Modal/RainbowKit (over Wagmi)/ in both lines.

### Nits (non-blocking)

- Because operate has no visible wallet UI, the `lightTheme({...})` configuration is essentially dead styling (the RainbowKit modal will never be opened from this app). Harmless as-is and keeping the same recipe across apps is reasonable — just calling it out.
- `accentColorForeground: 'white'` is new but a reasonable explicit default.
- Pre-existing-but-worth-noting: `apps/operate/common-util/config/wagmi.ts` doesn't set `ssr: true` (unlike `apps/launch`). Not introduced by this PR, but if the cross-app recipe should standardize on `ssr: true`, this would be the place to add it.

Otherwise LGTM.

mohandast52 (approved):
LGTM. Reviewed as part of the 7-app RainbowKit set; verified the shared concerns below.

- **Read-only app** — operate only uses `wagmiConfig` for `getPublicClient(...)` reads in `Contracts/hooks.ts`; no `useAccount`/ConnectButton/connect UI. So the migration is minimal (config + provider + dropping the boilerplate `w3m-button` JSX decl), and that's correct — no button to port.
- `getDefaultConfig` + `RainbowKitProvider` + `lightTheme` pattern matches the rest of the set.
- **Safe connector preserved**: verified in installed `@rainbow-me/rainbowkit@2.2.11` that `getDefaultConfig`'s default wallet list is `[safeWallet, rainbowWallet, coinbaseWallet, metaMaskWallet, walletConnectWallet]`, and `safeWallet` wraps `safe()` from `wagmi/connectors` — so the Safe-app flow is intact for the whole set.
- `tsc --noEmit` clean, `yarn nx lint operate` clean.

Approving. (Manual wallet smoke still worth doing across the set before the umbrella merges — modal open, theme parity, Safe-iframe connect.)

Tanya-atatakai (approved):
Re-reviewed. CLAUDE.md update matches the launch fix pattern cleanly. LGTM.

rajat2502 (approved):
LGTM. Smallest of the three; confirmed nothing in operate opens a wallet modal. Non-blocking: `RainbowKitProvider` + CSS ship for no functional gain — happy to keep for stack-consistency with launch/govern.

PR #410, opened 2026-05-22
chore(govern): migrate Web3Modal -> RainbowKit

atepem:
## Summary

Govern's RainbowKit migration. Stacked on **#406**; GitHub will auto-retarget to `precious/rainbowkit-migration` once #406 merges.

Same recipe as **#408** (launch), plus two govern-specific adjustments surfaced by wagmi 2.19's stricter typing under `getDefaultConfig`.

## Changes

| File | What |
|------|------|
| [common-util/config/wagmi.ts](apps/govern/common-util/config/wagmi.ts) | `defaultWagmiConfig` → `getDefaultConfig`. **Dropped the `SUPPORTED_CHAINS_WITH_RPCS` remap** — it lost the literal chain tuple that `getDefaultConfig`'s overload relies on. The `RPC_URLS[chain.id] \|\| chain.rpcUrls.default.http[0]` fallback is preserved inside the transports reducer instead. **This also fixes a latent bug** — see below. |
| [context/Web3ModalProvider.tsx](apps/govern/context/Web3ModalProvider.tsx) | `createWeb3Modal(...)` → `<RainbowKitProvider theme={lightTheme(...)}>` + RainbowKit CSS import. |
| [components/Login/LoginV2.tsx](apps/govern/components/Login/LoginV2.tsx) | `<w3m-button balance="hide" />` → `<ConnectButton.Custom>` rendering Ant `<Button>` for the three states. `resetState` + `clearUserState` wiring (govern-specific account-switch behavior) preserved as-is. |
| [index.d.ts](apps/govern/index.d.ts) | Dropped the `w3m-button` JSX intrinsic declaration. |
| [common-util/functions/requests.ts](apps/govern/common-util/functions/requests.ts) | Introduced `MAINNET_CHAIN_ID: number` alongside the existing literal `MAINNET = mainnet.id`. Wagmi 2.19's `simulateContract`/`readContracts` overloads reject the literal `1` against the new `wagmiConfig`'s chain tuple (`never` arg inference). Indexer expressions like `TOKENOMICS.addresses[MAINNET]` still need the literal form, so both are kept. |
| [CLAUDE.md](apps/govern/CLAUDE.md) | Wallet/Env lines updated: "Web3Modal (Wagmi)" → "RainbowKit (over Wagmi)". |

## Latent bugfix worth surfacing

The transports reducer fallback changed shape:

```diff
-  { [chain.id]: http(RPC_URLS[chain.id]) || chain.rpcUrls.default.http[0] }
+  { [chain.id]: http(RPC_URLS[chain.id] || chain.rpcUrls.default.http[0]) }
```

The old form's `||` operated on the **always-truthy** `Transport` object returned by `http(...)`, so the URL-string fallback was dead code. The previous `SUPPORTED_CHAINS_WITH_RPCS` remap (which mutated `chain.rpcUrls.default.http[0]`) was masking the bug because wagmi was still getting the right URL via the chain object's metadata. Collapsing both layers into the corrected `http(URL || fallback)` is a clean fix. Calling it out so it's not buried as a refactor — it's actually a fallback-path bugfix.

## MAINNET_CHAIN_ID tradeoff (documented)

The widening is unavoidable under wagmi 2.19's new chain-tuple inference, but the side-effect is that wagmi can no longer statically verify the chainId at the call sites that use `MAINNET_CHAIN_ID`. A typo or stale id would compile silently. Practical risk is low here (single mainnet-only file; addresses are still indexed via the literal `MAINNET`, so a mismatch would surface at the indexer layer). Centralizing the widening in two `const`s rather than scattering inline casts is the right call.

## Verified

- `yarn nx lint govern` passes
- `yarn nx build govern` passes

## Test plan

- [ ] CI passes
- [ ] Manual: pull, run dev server, exercise the connect/disconnect flow, then attempt a vote-weighting tx (the main `simulateContract` path that was rejected by the wagmi typing change)

### DIFF
--- a/apps/govern/CLAUDE.md
+++ b/apps/govern/CLAUDE.md
@@ -12,14 +12,14 @@ Olas governance: **voting**, **veOLAS**, **proposals**, **donations**, and **vot
 
 ## Stack
 
-- **Wallet**: Web3Modal (Wagmi). See `context/Web3ModalProvider.tsx`, `components/Login/`, `components/Contracts/`, `components/Donate/`, `components/Proposals/`, `components/VeOlas/`.
+- **Wallet**: RainbowKit (over Wagmi). See `context/Web3ModalProvider.tsx` (filename kept pending a cross-app rename), `components/Login/`, `components/Contracts/`, `components/Donate/`, `components/Proposals/`, `components/VeOlas/`.
 - **State**: Redux (`store/`).
 - **Key libs**: `util-constants`, `util-functions`, `util-contracts`, `ui-theme`, `ui-components`, `util-prohibited-data`, `common-contract-functions`, `common-middleware`, `util-ssr`.
 
 ## Env / backends
 
 - **Governor subgraph**: `NEXT_PUBLIC_GOVERNOR_SUBGRAPH_URL`
-- RPCs and Wallet Project ID for Web3Modal.
+- RPCs and Wallet Project ID for RainbowKit (WalletConnect Cloud projectId — required by `getDefaultConfig`).
 
 ## Structure
 
--- a/apps/govern/common-util/config/wagmi.ts
+++ b/apps/govern/common-util/config/wagmi.ts
@@ -1,3 +1,4 @@
+import { getDefaultConfig } from '@rainbow-me/rainbowkit';
 import { cookieStorage, createStorage, http } from 'wagmi';
 import {
   Chain,
@@ -13,7 +14,6 @@ import {
 } from 'wagmi/chains';
 
 import { RPC_URLS } from 'libs/util-constants/src';
-import { defaultWagmiConfig } from '@web3modal/wagmi';
 
 export const SUPPORTED_CHAINS: [Chain, ...Chain[]] = [
   mainnet,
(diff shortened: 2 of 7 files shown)


mohandast52 (approved):
LGTM. Part of the 7-app RainbowKit set.

- Standard `getDefaultConfig` + `RainbowKitProvider` + `ConnectButton.Custom` migration (replacing `defaultWagmiConfig` from web3modal).
- **Two good extras beyond the boilerplate:**
  - `requests.ts`: widening `chainId: MAINNET` (literal `1`) → `chainId: MAINNET_CHAIN_ID: number` across the params bags — needed because wagmi 2.19's stricter simulate/write/wait overloads reject the literal against the chain tuple. Correct fix; `tsc` confirms.
  - `wagmi.ts` transports: `http(RPC_URLS[id]) || chain.rpcUrls.default.http[0]` → `http(RPC_URLS[id] || chain.rpcUrls.default.http[0])`. The old form never fell back (a transport object is always truthy); the new form falls back the URL *inside* `http()`. Genuine bugfix.
- **Safe connector preserved** (verified `safeWallet` in `getDefaultConfig` defaults in rainbowkit 2.2.11). `cookieStorage` + `ssr: true` retained.
- `tsc --noEmit` clean, `yarn nx lint govern` clean.

Approving.

Tanya-atatakai (commented):
Reviewed against `precious/rainbowkit-migration`. The govern-specific bits (resetState/clearUserState wiring in `LoginV2.tsx`, `cookieStorage` + `ssr: true` in `wagmi.ts`) are all preserved, and there are no leftover `w3m-button` / `createWeb3Modal` / `@web3modal` / `W3M_BORDER_RADIUS` / `defaultWagmiConfig` references in `apps/govern/`.

### Findings

**1. `apps/govern/CLAUDE.md` wasn't updated** (same pattern as #408 / #409)

- L15: *"Wallet: Web3Modal (Wagmi). See `context/Web3ModalProvider.tsx`, …"*
- L22: *"RPCs and Wallet Project ID for Web3Modal."*

Root [CLAUDE.md](../blob/main/CLAUDE.md#app-level-claudemd-files) requires per-app docs to be updated on stack changes. s/Web3Modal/RainbowKit (over Wagmi)/.

**2. Silent latent-bugfix worth surfacing in the PR description** — `apps/govern/common-util/config/wagmi.ts`:

- Before: `{ [chain.id]: http(RPC_URLS[chain.id]) || chain.rpcUrls.default.http[0] }`
- After:  `{ [chain.id]: http(RPC_URLS[chain.id] || chain.rpcUrls.default.http[0]) }`

The old `||` operated on the always-truthy Transport object returned by `http(...)`, so the URL-string fallback was dead code. The previous `SUPPORTED_CHAINS_WITH_RPCS` remap (which mutated `chain.rpcUrls.default.http[0]`) was masking the bug because wagmi was still getting the right URL via the chain object's metadata. Collapsing both layers into the corrected `http(URL || fallback)` is a clean fix — nice catch. Suggest calling this out in the PR description so it's not buried as a refactor; it's actually a latent-bugfix.

**3. `MAINNET_CHAIN_ID: number` widening is a small type-safety regression** — `apps/govern/common-util/functions/requests.ts:32`

The widening is unavoidable (the inline comment correctly explains wagmi 2.19's `never` inference against the new chain tuple), but the side-effect is that wagmi can no longer statically verify the chainId at call sites — a typo or stale id would now compile silently. Practical risk is low here (single mainnet-only file; addresses are still indexed via the literal `MAINNET`, so a mismatch would surface at that layer). Centralizing the widening rather than scattering inline casts is the right call. Just worth flagging as a documented tradeoff in case the cross-app recipe needs to revisit it.

### Nits (non-blocking, same as #408 / #409)

- Provider filename `Web3ModalProvider.tsx` kept intentionally — fine.
- `borderRadius: 'small'` is a visual delta from the prior `1.4px` (RainbowKit theme accepts presets only).
- `accentColorForeground: 'white'` is new but a reasonable explicit default.

Otherwise LGTM.

Tanya-atatakai (approved):
Re-reviewed. CLAUDE.md update matches the launch fix pattern cleanly. LGTM.

rajat2502 (approved):
LGTM. The transports fallback bugfix (`http(URL || fallback)` vs the dead `http(URL) || fallback`) is a real cleanup and well-explained. `MAINNET_CHAIN_ID: number` widening is documented; longer-term we could drop the `[Chain, ...Chain[]]` annotation on `SUPPORTED_CHAINS` so wagmi infers the literal tuple, but that's umbrella-wide and out of scope here. Same umbrella-level WalletConnect-metadata / injected-wallet notes as #408 apply.

PR #412, opened 2026-05-22
chore(contribute): migrate Web3Modal -> RainbowKit

atepem:
## Summary

Contribute's RainbowKit migration. Stacked on **#406**; auto-retargets to `precious/rainbowkit-migration` once #406 merges.

**Contribute is the trickiest of the per-app PRs.** The existing config injects a custom Binance wallet connector (`@binance/w3w-wagmi-connector-v2`) that's not in RainbowKit's standard wallet registry — so the recipe diverges from launch in two ways.

## Changes

| File | What |
|------|------|
| [components/Login/config.ts](apps/contribute/components/Login/config.ts) | **Kept wagmi's `createConfig`** (instead of RainbowKit's `getDefaultConfig`, which doesn't accept a `connectors` override). RainbowKit's stock wallets are pulled in via `connectorsForWallets` (metaMask, rainbow, coinbase, walletConnect, injected, safe) and **concatenated with the existing `binanceConnector()`**. Preserves Binance wallet support. |
| [pages/_app.tsx](apps/contribute/pages/_app.tsx) | Inline `createWeb3Modal(...)` → `<RainbowKitProvider theme={lightTheme(...)}>` wrapping the ApolloProvider. RainbowKit CSS imported at top. |
| [components/Login/LoginV2.tsx](apps/contribute/components/Login/LoginV2.tsx) | `<w3m-button balance="hide" />` → `<ConnectButton.Custom>` rendering Ant `<Button>`. The "Your profile" Link button (rendered alongside the connect button when an address is set) is preserved as-is. |
| [index.d.ts](apps/contribute/index.d.ts) | Dropped the `w3m-button` JSX intrinsic. |

**Net diff:** 4 files, +79/-64.

## Verified

- `yarn nx lint contribute` passes
- `yarn nx build contribute` passes

## Test plan

- [ ] CI passes
- [ ] Manual: pull, run dev server, exercise both connect paths — standard RainbowKit modal (MetaMask/WalletConnect/etc.) AND the Binance wallet path (which uses its own injected provider)

### DIFF
--- a/apps/contribute/CLAUDE.md
+++ b/apps/contribute/CLAUDE.md
@@ -12,7 +12,7 @@ Contribute to Olas: staking, contributions, and profile. Integrates with a **bac
 
 ## Stack
 
-- **Wallet**: Yes. Web3Modal (Wagmi). See `pages/_app.tsx` (createWeb3Modal), `components/Login/`, `components/Staking/`, `components/Profile/`.
+- **Wallet**: Yes. RainbowKit (over Wagmi). See `pages/_app.tsx` (RainbowKitProvider), `components/Login/config.ts` (connectorsForWallets + custom Binance wallet), `components/Login/`, `components/Staking/`, `components/Profile/`.
 - **State**: Redux (`store/`) + Apollo Client for GraphQL where used.
 - **Key libs**: `util-functions`, `util-constants`, `ui-theme`, `ui-components`, `feature-service-status-info`, `common-middleware`, `util-ssr`.
 
@@ -23,7 +23,7 @@ Contribute to Olas: staking, contributions, and profile. Integrates with a **bac
 - **Discord**: `NEXT_PUBLIC_DISCORD_VERIFICATION_ADDRESS`
 - **AFMDB**: `NEXT_PUBLIC_AFMDB_URL`
 - **Agent DB** (server-side): `AGENT_DB_WALLET_PRIVATE_KEY`, `AGENT_TYPE_ID`, `ATTRIBUTE_ID_MAPPING`
-- RPCs and Wallet Project ID for Web3Modal.
+- RPCs and Wallet Project ID for RainbowKit / WalletConnect.
 
 ## Structure
 
--- a/apps/contribute/components/Login/LoginV2.tsx
+++ b/apps/contribute/components/Login/LoginV2.tsx
@@ -1,4 +1,5 @@
-import { Button } from 'antd';
+import { ConnectButton } from '@rainbow-me/rainbowkit';
+import { Button, Space } from 'antd';
 import Link from 'next/link';
 import { useEffect } from 'react';
 import { useDispatch } from 'react-redux';
@@ -152,7 +153,38 @@ export const LoginV2 = ({ onConnect: onConnectCb, onDisconnect: onDisconnectCb }
           </Link>
         </>
       )}
-      <w3m-button balance="hide" />
+      <ConnectButton.Custom>
+        {({ account, chain, openAccountModal, openChainModal, openConnectModal, mounted }) => {
+          if (!mounted) return null;
+
+          if (!account || !chain) {
+            return (
+              <Button type="primary" onClick={openConnectModal}>
+                Connect Wallet
+              </Button>
+            );
+          }
+
+          if (chain.unsupported) {
+            return (
+              <Button danger onClick={openChainModal}>
+                Wrong network
+              </Button>
+            );
+          }
+
+          return (
+            <Space size={4}>
+              <Button size="small" onClick={openChainModal}>
+                {chain.name}
+              </Button>
+              <Button size="small" onClick={openAccountModal}>
+                {account.displayName}
+              </Button>
+            </Space>
+          );
+        }}
+      </ConnectButton.Custom>
     </LoginContainer>
   );
 };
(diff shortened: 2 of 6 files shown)


mohandast52 (approved):
LGTM — the most involved of the set, handled well.

- Because contribute needs the **Binance** connector (not in RainbowKit's registry), it correctly uses the `connectorsForWallets` escape hatch instead of `getDefaultConfig` (which has no `connectors` override), then appends `binanceConnector()` to wagmi's `createConfig`. Sound approach, with a clear comment.
- **Safe preserved explicitly** — `safeWallet` is included in the `connectorsForWallets` list (the "Other" group), matching the Safe support `getDefaultConfig` gives the other apps. Good that this wasn't dropped.
- `ConnectButton.Custom` render + `RainbowKitProvider`/`lightTheme` in `_app.tsx`; removed the web3modal `excludeWalletIds`/`connectorImages` config. `cookieStorage` + `ssr` retained.
- `tsc --noEmit` clean, `yarn nx lint contribute` clean.

One non-blocking nit: `apps/contribute/CLAUDE.md` still says **"Wallet: Web3Modal (Wagmi) ... (createWeb3Modal)"** — please update it to RainbowKit per the repo's keep-app-CLAUDE.md-accurate convention.

Approving. Manual smoke should include the Binance wallet path specifically.

Tanya-atatakai (changes requested):
Reviewed against `precious/rainbowkit-migration`. The decision to keep wagmi's `createConfig` and use `connectorsForWallets` instead of `getDefaultConfig` is the right escape hatch for the custom Binance W3W connector — well-judged. But I think the migration drops the Binance connect path in a way that's likely a regression, so leaving as request-changes pending verification.

### Concerns

**1. Binance connector is no longer reachable from the modal UI (likely blocker)**

In `apps/contribute/components/Login/config.ts:45`, `binanceConnector()` is appended to wagmi's `connectors` array — but RainbowKit's modal only renders wallets declared via `connectorsForWallets`. Binance is *not* in that list. I grepped the rest of the app and there's no `useConnect({ connector: binanceConnector })` or any other UI entry point that triggers the connector by reference.

The previous Web3Modal config had:
```ts
connectorImages: {
  ['wallet.binance.com']: '/images/binance-wallet.svg',
},
```
which made Binance clickable inside the Web3Modal modal. That UX path is gone.

The PR description says "exercise the Binance wallet path (which uses its own injected provider)" — but `@binance/w3w-wagmi-connector-v2` is a WaaS QR/SDK connector, not an injected provider, so it won't auto-surface via RK's "Detected" wallet group either.

**To fix**, the Binance wallet needs to be wrapped as a RainbowKit-compatible wallet entry (a `Wallet` object with `id`, `name`, `iconUrl`, and `createConnector`) and included in the `connectorsForWallets` array. Something like:

```ts
const binanceWallet = (): Wallet => ({
  id: 'binance-w3w',
  name: 'Binance Wallet',
  iconUrl: '/images/binance-wallet.svg',
  iconBackground: '#F0B90B',
  createConnector: () => binanceConnector(),
});

// then add `binanceWallet` to one of the `wallets:` arrays
```

Please verify the Binance flow end-to-end against the previous behavior before merging.

**2. `excludeWalletIds` dropped (minor)**

The old config excluded two WalletConnect IDs:
```
'8a0ee50d1f22f6651afcae7eb4253e52a3310b90af5daef78a8c4929a9bb99d4'
'cbc11415130d01316513f735eac34fd1ad7a5d40a993bbb6772d2c02eeef3df8'
```
Probably a no-op in practice (neither is in RK's curated registry, and RK doesn't filter WC peers), but worth a quick check whether those exclusions were load-bearing. If they were just stale, fine to drop silently.

**3. `apps/contribute/CLAUDE.md:15` wasn't updated**

Still reads *"Wallet: Yes. Web3Modal (Wagmi). See `pages/_app.tsx` (createWeb3Modal), …"*. s/Web3Modal/RainbowKit (over Wagmi)/, drop the `createWeb3Modal` reference.

### Nits

- `borderRadius: 'small'` is a visual delta from the prior `'0.7125px'` (RainbowKit only accepts presets — `'0.7125px'` actually rounds down to `none` visually for many components anyway).
- "Your profile" Link button preservation correctly verified — still rendered alongside the connect button when an address is set.

Apart from the Binance regression, the migration is well-scoped.

Tanya-atatakai (approved):
Re-reviewed both follow-up commits.

**`fix(contribute): wrap Binance connector as RainbowKit wallet entry`** — Binance is now a proper `Wallet` object (id, name, iconUrl, iconBackground, createConnector) declared in the "Other" group alongside `safeWallet`. `binanceConnectorFn = getWagmiConnectorV2()` is hoisted once and reused via `createConnector`, and the duplicate `binanceConnector()` was correctly removed from the wagmi connectors array — `connectors` now flows entirely from `connectorsForWallets`. Resolves the blocker.

**`chore(contribute): add WalletConnect metadata parity`** — nice catch I hadn't flagged: the old Web3Modal config passed `description`, `url`, `icons` in WalletConnect metadata (these surface during pairing approval), and the initial migration only carried `appName` + `projectId`. Restoring `appDescription` / `appUrl` / `appIcon` (with the same Olas GitHub avatar URL) preserves UX parity for users approving WC sessions.

CLAUDE.md updated to point at the new `connectorsForWallets + custom Binance wallet` setup. LGTM.

PR #415, opened 2026-05-22
chore(marketplace): migrate Web3Modal -> RainbowKit

atepem:
## Summary

Final per-app PR in the rainbowkit migration. Completes the 7-app batch on the `precious/rainbowkit-migration` umbrella.

Marketplace is the most complex case in this batch: inline `createWeb3Modal` in `pages/_app.tsx`, a `.jsx` LoginV2 that mixes a responsive web3modal button with an app-specific "Switch network" YellowButton, AND a Solana branch.

## Changes

| File | What |
|------|------|
| [common-util/Login/config.tsx](apps/marketplace/common-util/Login/config.tsx) | `defaultWagmiConfig` → `getDefaultConfig`. **Dropped the in-config `rpcUrls` remap** — same lesson as govern's PR (the remap loses the literal chain tuple `getDefaultConfig`'s overload needs, leading to `never` arg inference downstream). The `RPC_URLS[chain.id] \|\| chain.rpcUrls.default.http[0]` fallback is preserved by threading it through the `transports` reducer instead. `appName` set from existing `wagmiMetadata.name`. Solana/SVM sections untouched. |
| [pages/_app.tsx](apps/marketplace/pages/_app.tsx) | Inline `createWeb3Modal(...)` → `<RainbowKitProvider theme={lightTheme(...)}>` wrapping `<Layout>` inside the existing `QueryClientProvider` / `WagmiProvider` tree. RainbowKit CSS imported at top. `W3M_BORDER_RADIUS` import dropped. |
| [common-util/Login/LoginV2.jsx](apps/marketplace/common-util/Login/LoginV2.jsx) | `<w3m-button balance={screens.xs ? 'hide' : 'show'} />` → `<ConnectButton.Custom>` with the same responsive behavior (mobile hides balance, desktop prefixes balance + symbol to address pill). **Existing YellowButton "Switch network" path** (driven by `isConnectedToWrongNetwork`) **AND the `SolanaWallet` SVM branch are preserved as-is.** |

**Net diff:** 3 files, +66/-33.

No `index.d.ts` edit — marketplace's index.d.ts didn't declare the `w3m-button` JSX intrinsic (LoginV2 is `.jsx`, so the intrinsic was implicitly typed as `any`).

## Verified

- `yarn nx lint marketplace` passes
- `yarn nx build marketplace` passes

## Completes the rainbowkit batch

All 7 in-scope apps now have rainbowkit PRs on the umbrella:
- #408 launch
- #409 operate
- #410 govern
- #411 build
[...]

PR #413, opened 2026-05-22
chore(bond): migrate Web3Modal -> RainbowKit

atepem:
## Summary

Final per-app PR in the in-scope batch (originally — marketplace's own RainbowKit migration ended up in a separate PR after #406 merged). Stacked on **#406**; auto-retargets to `precious/rainbowkit-migration` once #406 merges.

**Bond is the trickiest UI-wise** — a `.jsx` app with a responsive connect button that historically showed BOTH a network selector AND the account/balance button side-by-side (`<w3m-network-button>` + `<w3m-button balance={screens.xs ? 'hide' : 'show'}>`).

## Changes

| File | What |
|------|------|
| [common-util/config/wagmi.js](apps/bond/common-util/config/wagmi.js) | `createConfig` + raw connectors → `getDefaultConfig`. **Also a deliberate dead-config cleanup** — see below. |
| [context/Web3ModalProvider.jsx](apps/bond/context/Web3ModalProvider.jsx) | `createWeb3Modal(...)` → `<RainbowKitProvider theme={lightTheme(...)}>`. Standard recipe. |
| [common-util/Login/LoginV2.jsx](apps/bond/common-util/Login/LoginV2.jsx) | `<w3m-network-button>` + `<w3m-button balance={screens.xs ? 'hide' : 'show'} />` → a single `<ConnectButton.Custom>` rendering a chain pill + account pill via Ant `<Button>`. **Responsive behavior preserved**: on `xs` the balance is hidden, on larger breakpoints the formatted balance + symbol is prefixed to the account display. `watchAccount` / `setUserBalance` / `isAddressProhibited` wiring kept as-is. |
| [CLAUDE.md](apps/bond/CLAUDE.md) | Wallet stack line updated: "Web3Modal (EVM)" → "RainbowKit (over Wagmi, EVM)". Solana adapter half unchanged. |

**Net diff:** 3 files + CLAUDE.md.

**No `index.d.ts` change** — bond's `index.d.ts` didn't declare the `w3m-button` JSX intrinsic. LoginV2 is `.jsx`, so the web-component intrinsic was implicitly typed as `any`.

## Dead-config cleanup (worth surfacing)

Two wagmi-v1 leftovers removed from the config in the same edit, calling out so a future archaeologist knows they're deliberate dead-config removals, not behavior changes:

[...]

PR #411, opened 2026-05-22
chore(build): migrate Web3Modal -> RainbowKit

atepem:
## Summary

Build app's RainbowKit migration. Stacked on **#406**; auto-retargets to `precious/rainbowkit-migration` once #406 merges.

Same recipe as **#408** (launch), plus deletion of the build app's standalone `web3modal.d.ts` file (a separate ambient module that only existed to declare the `w3m-button` JSX intrinsic — no longer needed).

## Changes

| File | What |
|------|------|
| [components/Login/config.ts](apps/build/components/Login/config.ts) | `createConfig` + raw connectors → `getDefaultConfig`. Connector list (injected, walletConnect, coinbase) is auto-bundled by RainbowKit. Standalone `projectId` export left in place. |
| [context/web3ModalProvider.tsx](apps/build/context/web3ModalProvider.tsx) | `createWeb3Modal(...)` → `<RainbowKitProvider theme={lightTheme(...)}>` + RainbowKit CSS import. |
| [components/Login/LoginV2.tsx](apps/build/components/Login/LoginV2.tsx) | `<w3m-button balance="hide" />` → `<ConnectButton.Custom>` rendering Ant `<Button>`. Redux-store chainId sync (`setChainId`) + `window.MODAL_PROVIDER` wiring (build-specific connector glue) preserved as-is. |
| ~~web3modal.d.ts~~ | **Deleted.** Single-purpose ambient module — only declared the `w3m-button` JSX intrinsic. |

**Net diff:** 4 files (3 modified, 1 deleted), +52/-54.

## Verified

- `yarn nx build build` passes
- `yarn nx lint build` is pre-existing broken on the umbrella (`Invalid lint configuration` — config issue unrelated to this PR, present on `precious/rainbowkit-prereq-deps` too)

## Test plan

- [ ] CI passes (build)
- [ ] Manual: pull, run dev server, exercise connect flow + staking contract creation

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #416, opened 2026-05-26
feat: migrate web3.js -> viem and Web3Modal -> RainbowKit across 7 apps

atepem:
## Summary

Two coupled migrations across all 7 frontend apps, bundled as one rollup since the rainbowkit umbrella was branched off the viem umbrella and synced with it:

1. **web3.js → viem** — replaces the old `web3` package with viem/wagmi v2 actions (`readContract`, `simulateContract`, `writeContract`, `waitForTransactionReceipt`, `multicall`, `getPublicClient`, etc.) across all 7 apps. The marketplace 3rdStep multisig flow was the last holdout and ships here too. Marketplace is now `web3.js`-free.
2. **Web3Modal → RainbowKit** — swaps `@web3modal/wagmi` for `@rainbow-me/rainbowkit` 2.2.11. Per-app config + provider + LoginV2 rewrites. Each app uses `<ConnectButton.Custom>` rendering Ant Design buttons for the three states (disconnected / wrong-network / connected with chain + address pills).

Plus the prereq dep bump (wagmi 2.5.20 → 2.19.5, `@wagmi/core` 2.6.17 → 2.22.1, viem 2.37.5 → 2.50.4, `@rainbow-me/rainbowkit` 2.2.11 added) and a security pin (`js-cookie` 3.0.7 to clear GHSA-qjx8-664m-686j).

## What's in this rollup

Already-landed sub-PRs (in chronological merge order):

**Viem migration:**
- #395 launch + operate (dead-file deletion)
- #397 build
- #399 govern
- #400 contribute
- #402 bond
- #403 marketplace
- #407 marketplace 3rdStep multisig flow

**RainbowKit migration:**
- #406 prereq deps (wagmi/viem/rainbowkit bumps + tsconfig es2022 + js-cookie 3.0.7 security pin)
- #408 launch
- #409 operate
- #410 govern (also a latent transports-fallback bugfix surfaced)
- #411 build
- #412 contribute (Binance wallet wrapped as a RainbowKit Wallet entry to preserve support)
- #413 bond (wagmi-v1 leftover cleanup: `autoConnect`, `options.rpc`)
- #415 marketplace

Plus on the umbrella:
- Eslint-delete wrapper removed from each app's `next.config.js` (Vercel modifyConfig was crashing on `path.join(undefined)` post-delete)
- Explicit `pages/404.tsx` added to 6 apps (Vercel's chmod step needed it under `static/<locale>/`)
- `@web3modal/wagmi` dropped from root `package.json` (now unused)
- `qr@0.5.5` pinned in `resolutions` to fix RainbowKit's WalletConnect QR `invalid border=0` crash (cuer passes `border: 0`; qr@0.6.0 added stricter `<= 0` validation — pin holds the transitive at 0.5.5)
- Duplicate chain dropdown removed from launch + marketplace `LoginV2` (Layout's `SwitchNetworkSelect` already covers it); connect-button section bumped to `size="large"` to match the 40px Select height
- "Confirm in wallet" indicator on the shared `SwitchNetworkButton` while a chain switch is pending (mirrors RainbowKit's chain-modal indicator)

## Post-review fixes (latest round)

Addressed in commits `ffe2c76a`, `35498886`, `d5b9fe4b`, `ab2fece7`:

**Must-fix bugs (HIGH):**
- `isHashApproved` polling (marketplace step 3) — added 5-minute timeout, optional `AbortSignal`, and a single `cleanup()` helper that's called from every exit path. Previously the setInterval leaked and the parent `await` hung forever if the chain stalled or the event was dropped.
- `getContractEvents` `fromBlock: 0n` (marketplace step 3) — bounded to `startingBlock - 9000` so the request stays under Alchemy's 10k-block window. Previously the wide query was rejected by most public RPCs, the exception was swallowed by the surrounding try/catch into a `console.error`, and the user saw no error state.
- Contribute multicall (`getLatestMintedNft`) — chunked the `ownerOf` calls at 100 per batch. A single multicall over thousands of tokens exceeded Multicall3's ~30M-gas budget or the RPC's max payload; with `allowFailure: true` the failure returned empty strings for every entry, making "owned NFT" silently look like "no NFT minted."

**Should-fix consistency (MED):**
- SSR connection-state persistence now wired in every app. `bond`, `build`, `launch`, `operate` were missing both `cookieStorage` and `cookieToInitialState` — wallet was lost on every full-page reload. `govern` had `cookieStorage` configured but didn't read it. Now every app passes `cookieToInitialState(wagmiConfig)` into its provider's `initialState` prop, matching the `marketplace`/`contribute` pattern.
- `isMounted` defer (for the RainbowKit + WC first-click "invalid border=0" race) is now applied uniformly across all 7 apps. `launch`'s previously over-broad gate (which deferred the entire `Layout`+page tree) is narrowed to just the wallet provider — Layout still SSRs, only wallet-aware children wait for hydration.
- Safe-via-WC redirect suppression lifted into a shared `useSuppressSafeWcRedirect` hook in `libs/util-functions`. Tightened from `includes` to `startsWith` (less risk of suppressing benign URLs that happen to contain the substring). Cleanup is ref-checked — only restores `window.open` if it still equals the patched function, so any sibling library that wraps a layered patch on top isn't silently un-wrapped. Applied to `govern` as well (Safe-via-WC is supported there per the smoke matrix below).
- Explicit RPC fallback `http(RPC_URLS[chain.id] || chain.rpcUrls.default.http[0])` in the apps that were missing it, matching `govern`/`marketplace`/`contribute`.

**Docs:**
- Root `CLAUDE.md` "Web3 Integration" / "Key Dependencies" sections updated from Web3Modal / wagmi 2.5 / viem 2.37 to RainbowKit / wagmi 2.19 / viem 2.50 plus the rainbowkit version.

## Screenshots

| Before | After |
|---|---|
| Screenshot 2026-05-26 150807 | Screenshot 2026-05-26 150817 |
| Screenshot 2026-05-26 151144 | Screenshot 2026-05-26 151134 |

And walletconnect further wallets:
Screenshot 2026-05-26 151330

## Verified (automated)

- All per-app PRs already reviewed and merged into this umbrella
- `yarn install --frozen-lockfile`, `yarn lint:lockfile`, `yarn audit:install-hooks`, `yarn audit:prod` all clean
- `yarn nx build` passes for all 7 apps (most recently rerun after the post-review fixes)
- `yarn nx lint` passes for 6 apps (the `build` app's lint config has a pre-existing config issue unrelated to this work)
- Vercel preview deploys green across all 7 apps on the umbrella

## Verified (manual smoke against Vercel previews)

The decoded contract reverts below are the **target signal**, not failures. They prove the full chain (RainbowKit → wagmi v2.19 → viem 2.50 ABI error decoder → MetaMask signing surface → on-chain call) is wired correctly. Undecoded `0x…` blobs would be a problem; named errors with typed args are not.

### Cross-cutting (covered indirectly via per-app tests)
- Connect + disconnect with MetaMask
- RainbowKit modal opens cleanly (no `invalid border=0` crash post-`qr@0.5.5` pin)
- Button height + duplicate-pill removal — visually confirmed via before/after screenshots

### Per-app

| App | Read | Write | Signal | Notes |
|---|---|---|---|---|
| **bond** | ✅ | ⚠️ not exercised | session persists across refresh; responsive button works at <480px | The bonding-products read hit a pre-existing `ORCA_WHIRLPOOLS` subgraph error before we reached a write attempt — outside the migration scope, will be retested on staging |
| **build** | ✅ | ✅ | claim-rewards → decoded `OwnerOnly(0xD19f…, 0x5E07…)` | |
| **contribute** | ✅ | ✅ | wallet modal includes Binance Wallet (custom `connectorsForWallets` recipe preserved); Binance WC QR renders cleanly | Profile page infinite-skeleton is pre-existing, unrelated to migration |
| **govern (EOA)** | ✅ | ✅ | vote → decoded `LockExpired(account=0x5E07…, deadline=0, curTime=1779926400)` | Validates the `MAINNET_CHAIN_ID: number` widening for wagmi 2.19's stricter chain-tuple typing |
| **govern (Safe-via-WC)** | ✅ | ✅ | Safe address `0x0df3…D37A` in pill; `voteForNomineeWeightsBatch` → decoded `LockExpired` with **Safe as both `sender` and `account`** | Chain switch correctly routes through Safe's WC popup (Safes are chain-locked); session persists across refresh; clean disconnect from the address-pill modal returns to Connect-Wallet state. Safe-as-App (iframe + manifest) path was never supported — no `manifest.json` shipped, pre-existing, out of scope. |
| **launch** | ✅ | ✅ | create-staking → `0x7c946ed7` (ImplementationNotWhitelisted — expected for unauthorized impls on mainnet) | |
| **marketplace** | ✅ | ✅ | mint with threshold=2 → decoded `WrongThreshold(currentThreshold=2, minThreshold=3, maxThreshold=3)`; with threshold=3 → clean MetaMask simulation showing `You receive: #59` + gas (0.0013 ETH) | |
| **operate** | ✅ | ⚠️ not exercised | connect/disconnect + chain switch across supported chains | No write path was exercised — covered by structural CI gates (build, lint, audit) but not by a runtime signal. Will be retested on staging. |

### Deferred verifications

- **Marketplace step 3 multisig update flow** (#407 code path) — requires connecting via a Gnosis Safe with funds; will verify post-merge on staging. The timeout + abort + cleanup wiring added in this round should make any future stall surface as a 5-min `Timed out` rejection instead of an indefinite hang.
- **SVM/Solana write flows** in marketplace — requires testnet SOL
- **bond + operate write paths** — bond blocked by a pre-existing subgraph error; operate not exercised. Both will be retested on staging.
- **Post-broadcast UI flows** (toast wiring, refetch invalidation, multi-step wizard advancement on receipt) — every smoke path either reverted at simulate or stopped at the MetaMask popup, so the broadcast → receipt → UI-update layer is unverified by this round. Same viem core as before; structurally unchanged by the migration. Would be the candidate for a Tenderly fork test if there's a regression on staging.

## Known follow-ups (separate PRs, non-blocking)

- **CSP `connect-src` additions**: RainbowKit 2.2 / Reown's appkit pings `https://pulse.walletconnect.org` (telemetry) and `https://api.web3modal.org/appkit/v1/config` (remote config). Neither is in any app's `next-safe` `connect-src` directive, so the browser blocks them and Reown falls back to local defaults. Non-blocking — apps function normally — but worth allowlisting to clear the console pollution.
- Rename `Web3ModalProvider.tsx` → `WalletProvider.tsx` across all apps (filename kept here to minimize migration churn)
- Fix `launch`/`govern`/`marketplace` serve targets in their `project.json` (broken since the b40b15a7 webpack switch — local `nx run <app>:serve` still throws `paths[1] must be of type string`)
- Contribute profile page infinite-skeleton when staking contracts are temporarily unavailable — pre-existing behavior, unrelated to this migration
- Safe-App (iframe-embedded) support — ship per-app `manifest.json` + icon assets + verify CSP `frame-ancestors` allows Safe's domain. Separate feature ticket if there's product demand.
- Restore the 6 marketplace test suites currently `describe.skip`'d with `TODO(viem-migration)` (`ListAgents/mint`, `ListAgents/utils`, `ListComponents/mint`, `ListComponents/utils`, `ListServices/mint`, `ListServices/utils`) — the viem migration changed contract-interaction shapes; mocks need rewriting. Whoever picks this up should add explicit coverage for `bigint`-bond accumulation (the path where the ca3af6ee5 fix landed).
- Migrate `useAccount({ onConnect, onDisconnect })` callsites to `useAccountEffect` in bond/build/marketplace/govern. The wagmi v2 options arg doesn't accept those callbacks (silently no-op); functionally compensated by parallel `useEffect`s but the dead code is misleading.
- Marketplace `requireChainId()` not-connected branches are dead code — `getChainId()` defaults to mainnet before ever returning null, so reads on first page load before chain is set in sessionStorage silently hit mainnet even if the URL says `/gnosis/`. UX bug worth a focused fix.
- Marketplace `ListServices/mint.jsx` and `update.jsx` pass raw strings for `uint32[]` / `uint32` args; `ListAgents` and `ListComponents` use `.map(BigInt)`. Viem coerces, so functionally OK, but the inconsistency is a footgun.
- `build`'s `getOwnerIncentives` destructure-as-tuple → consumer re-accesses by name; commit to one style.
- Drop dead root dep `web3: 1.10.4` (no source imports remain post-migration). Out of scope for this PR; supply-chain-surface cleanup.


apps/marketplace/components/ListServices/ServiceState/utils.jsx
```diff
@@ -37,33 +45,35 @@ export const getBonds = async (id, tableDataSource) => {
      * slotsArray = [2, 3]
      * bondsArray = [2000, 4000]
      */
-
     const { bond, slots } = response.agentParams[i];
     slotsArray.push(slots);
     bondsArray.push(bond);
```
Tanya-atatakai, line 51: 🔴 HIGH: `getAgentParams.bond` is uint96, so viem returns **bigint** here (web3 returned a string). This flows into `transformSlotsAndBonds` (`helpers/functions.jsx:68`), which does `totalBonds += numberOfAgentAddress * bond;` — `number * bigint` throws `TypeError: Cannot mix BigInt and other types`. That helper isn't in this PR's diff, so the type change wasn't propagated. `totalBonds` ultimately feeds the `registerAgents` ETH `value`. Fix by accumulating as bigint (`BigInt(numberOfAgentAddress) * bond`), not `Number(bond)` (precision loss above 2^53).

apps/marketplace/components/ListServices/ServiceState/utils.jsx
```diff
@@ -226,53 +216,72 @@ export const onStep2RegisterAgents = async ({
...
+  const chainId = requireChainId();
   return Promise.all(
-    (source || []).map(async ({ agentId }) => {
-      const bond = await contract.methods.getAgentBond(id, agentId).call();
-      return bond;
-    }),
+    (source || []).map(({ agentId }) =>
+      readContract(wagmiConfig, {
+        ...serviceRegistryTokenUtilityParams(chainId),
+        functionName: 'getAgentBond',
```
Tanya-atatakai, line 249: 🔴 HIGH: `getAgentBond` is uint256, so each element here is **bigint** (web3 returned a string). The consumer `2StepActiveRegistration/index.jsx:120` does `totalTokenBonds += addressCount * bond;` in the **render body** → `number * bigint` `TypeError`, crashing the page for non-ETH custom-token services in Active Registration. That file isn't in this PR's diff. The result also feeds `amountToApprove`. Fix by accumulating as bigint, not `Number()`.

Tanya-atatakai (changes requested):
## Review: RainbowKit + viem migration

The migration is broadly solid and carefully done — return shapes (array vs object), bigint/number handling, units, payable `value`, bytes32 args, and event-log decoding all check out across govern/build/contribute, and the marketplace `getPublicClient`/`writeContract` wiring + transports preserve the keyed RPC URLs (no public-RPC fallback). Lint and tests pass for all 7 apps.

However there are **2 verified runtime crashes** that block merge, plus a consistency risk. Details below.

---

### 🔴 HIGH — verified `number * bigint` crashes (must fix)

The reads migrated from web3.js (returns **string**) to viem (returns **bigint**), but two *consumer* files that compute `number * bond` were **not updated** — and they are **not part of this PR's diff**, which is why they slipped through. `number * bigint` throws `TypeError: Cannot mix BigInt and other types`. Under web3 the string coerced numerically. I've anchored inline comments to the migrated lines that now produce the bigint; the actual crash sites are:

- **`apps/marketplace/components/ListServices/ServiceState/2StepActiveRegistration/index.jsx:120`** — `totalTokenBonds += addressCount * bond;` runs in the **render body**, so a non-ETH custom-token service in Active Registration **crashes on render**. Result feeds `amountToApprove` (line 136). Source: `getTokenBondRequest` → `getAgentBond` (uint256 → bigint).
- **`apps/marketplace/components/ListServices/helpers/functions.jsx:68`** — `totalBonds += numberOfAgentAddress * bond;` inside `transformSlotsAndBonds`. Wrapped in a try/catch, so it surfaces as "Error while fetching bonds" rather than a hard crash, but bond calculation is fully broken and `totalBonds` later feeds the `registerAgents` ETH `value`. Source: `getBonds` → `getAgentParams.bond` (uint96 → bigint).

**Suggested fix (both):** accumulate as bigint — not `Number(bond)`, since these are wei/token amounts feeding approvals and `value` and would lose precision above 2^53:
```js
let totalTokenBonds = 0n;
totalTokenBonds += BigInt(addressCount) * bond;
// keep `${totalTokenBonds}` / .toString() as before
```

---

### 🟡 MED — provider-mount defer applied inconsistently

Commit `165a9d8b` fixes the WalletConnect first-click crash (`invalid border=0`) by deferring the wallet provider behind `isMounted`, framing it as a general RainbowKit+WalletConnect race. But only **contribute** and **launch** defer; **bond, build, govern, operate, marketplace** mount eagerly and may still hit the race on first connect. Either apply the defer uniformly or manually confirm the other 5 don't reproduce it. (Also: launch gates the entire Layout+page tree, not just the provider — heavier SSR trade-off than necessary.)

---

### 🟢 LOW — informational

- build/contribute reads use the legacy `getChainId()` provider-sniffing helper rather than wagmi's connected chain (pre-existing, not a regression — worth a follow-up).
- marketplace `requireChainId()` (`common-util/functions/index.ts:51`): the not-connected guard branches are dead code; reads silently default to mainnet if session storage is empty.
- marketplace `ListServices/{mint,update}.jsx` pass raw strings for uint array/threshold args; works only because viem coerces internally — inconsistent with `ListAgents` which uses `.map(BigInt)`.

mohandast52:
## End-to-end review

Checked out the head locally, verified each prior reviewer's claims against the actual code, and did an independent pass on the migration mechanics.

### Verdict

There are **2 unfixed correctness issues** that produce real user-visible failures, plus an SSR-persistence regression that's silently inherited. The two HIGH bigint crashes from #416 (review by @Tanya-atatakai) are fixed in `ca3af6ee5`. Other findings below remain open and I found additional ones.

Recommend: address the **must-fix** block, ship the **should-fix** block as part of this PR (cheap), defer the rest to follow-ups.

---

### 🔴 Must-fix before merge

**1. `isHashApproved` polling has no timeout *and* no React cleanup**
`apps/marketplace/components/ListServices/ServiceState/3rdStepFinishedRegistration/helpers.jsx:16-39`

The `setInterval` fires every 5s with no max-wait and no abort signal. If the chain stalls, the RPC drops the event, or the user navigates away mid-flow:
- the Promise never settles → the parent `await isHashApproved(...)` hangs forever (line 189 in `utils.jsx`), and
- the interval keeps firing in the background (memory leak; new requests every 5s).

This is the Safe step-3 update-multisig path (a write flow with real money queued behind it) — silent hang is the worst possible UX.

**Fix:** add a max timeout (5 min is reasonable) AND accept an AbortSignal/cancel handle that the caller wires to component unmount.

**2. `getContractEvents` with `fromBlock: 0n` will fail on most chains**
`apps/marketplace/components/ListServices/ServiceState/3rdStepFinishedRegistration/utils.jsx:164-172`

`eth_getLogs` with a fromBlock of 0 over a 20M-block chain is rejected by Alchemy (10k-block window), Infura, and most public RPCs. The exception will throw out of the surrounding try/catch and only `console.error` it — the user sees nothing, and the flow silently breaks.

This isn't new behavior in spirit (web3.js `getPastEvents({fromBlock:0})` had the same problem), but the migration is the right time to fix it. Bound the scan to a sensible window (e.g. last ~9k blocks to stay under Alchemy's `eth_getLogs` cap, or use the same `startingBlock - N` pattern that `isHashApproved` uses).

**3. Contribute `multicall` can exceed gas/payload limits**
`apps/contribute/common-util/api/nftDetails.ts:41-52`

The migration replaces N serial `await contract.methods.ownerOf(i).call()` with a single `multicall({contracts: [...]})` containing one entry per token from `1n` to `totalSupply`. For a contribute-badges contract with thousands of mints, this single eth_call will exceed Multicall3's gas budget (~30M) or the RPC's max payload — and `allowFailure: true` silently returns empty addresses, so the **lookup will return \"no NFT minted\"** when the user actually has one.

**Fix:** chunk into batches of ~100 calls and concatenate results. Or — better, since the function name is `getLatestMintedNft` — query `Transfer(to=account)` event logs and grab the most recent. (Note: the pre-existing `findIndex` returns the *first* match, not the latest, so the function name has been a lie for a while. Pre-existing, not introduced here, but worth fixing in the same pass.)

---

### 🟡 Should-fix in this PR (small, related, cheap)

**4. SSR connection-state persistence is half-configured**

After the migration, the wagmi configs look like this:

| App | `cookieStorage` in config | `cookieToInitialState` in provider | Status |
|---|---|---|---|
| marketplace | ✅ | ✅ | correct |
| contribute  | ✅ | ✅ | correct |
| govern      | ✅ | ❌ | **broken**: cookies written but never read on SSR |
| bond, build, launch, operate | ❌ | ❌ | preserves pre-PR brokenness |

`ssr: true` is set in every app but only marketplace and contribute actually wire SSR hydration end-to-end. The result for the other 5 (especially **launch** and **operate**, which are 8-chain apps with SSR): connection state is lost on every full-page reload, and React 19's hydration check will flag the mismatch.

Govern is the weirdest case — it has `cookieStorage` configured (`apps/govern/common-util/config/wagmi.ts:34`) but the provider at `apps/govern/context/Web3ModalProvider.tsx` doesn't pass `initialState`. So cookies *are* being written for no reason. Either complete the wiring or remove `cookieStorage` to be honest about it.

The bond/build/launch/operate omission was pre-existing, but every wagmi config in those apps was touched in this PR — fixing it now is two extra lines per app. Coding-agent flagged this; I confirmed it.

**5. Provider-mount defer applied inconsistently (Tanya's MED, still open)**

Only `apps/contribute/pages/_app.tsx:42-46` and `apps/launch/pages/_app.tsx:31-34` defer the provider behind `isMounted`. The other 5 apps mount eagerly, so the WC \"invalid border=0\" first-click race that motivated the workaround is still live in bond/build/govern/operate/marketplace.

Two options:
- Apply the defer uniformly (lift the pattern into a shared provider component to avoid drift).
- Confirm the other 5 don't reproduce the race and add a comment explaining why launch + contribute are special.

Note also: launch gates the *entire* `<Layout>`+page tree behind `isMounted`, not just the provider. That's a needless SSR sacrifice — only the wallet-aware children need to wait. Worth narrowing the gate to just `<Web3ModalProvider>` (and letting Layout render on the server).

**6. `launch` `window.open` monkey-patch has a layered-patch leak**
`apps/launch/pages/_app.tsx:42-53`

The cleanup restores `window.open` to whatever `originalOpen` was *at mount time*. If any other library or effect wraps `window.open` *after* this effect mounts but before it unmounts, the unmount cleanup will restore to the pre-mount value — silently un-wrapping the other library's patch. The inline comment claims it \"works correctly with layered patches\" — it doesn't; it works correctly only with React strict-mode's double-invoke (because both invocations save the same `originalOpen`).

Safer pattern: store the function on a `ref` and check `window.open === ourPatched` before restoring. Also, the `url.includes('app.safe.global')` substring match is too loose — `url.startsWith('https://app.safe.global')` would be tighter and avoids matching e.g. a query param.

Also: Safe-via-WC is also supported in **govern** (the PR description tests it explicitly). The same redirect-tab annoyance applies there. Either lift this patch to a shared place and apply uniformly, or scope it tightly so it's clearly intentionally launch-only.

---

### 🟢 Notes / follow-ups (not blocking)

- **6 marketplace test suites are `describe.skip`'d** with `TODO(viem-migration)` (`apps/marketplace/tests/components/{ListAgents,ListComponents,ListServices}/`). These cover mint, update, and utils — *exactly* the paths where the bigint*number crash happened. The PR's structural CI gates (build/lint/audit) caught zero of those bugs. Whoever picks up the test rewrite should make sure the bigint-bond accumulation has explicit coverage.
- **wagmi v2 `useAccount({ onConnect, onDisconnect })` is dead code.** In wagmi v2 the options arg doesn't take callbacks (moved to `useAccountEffect`). All four apps that still use the v1 shape (`apps/bond/common-util/Login/LoginV2.jsx:35-50`, build, marketplace, govern) silently no-op those callbacks. Functionally compensated by parallel useEffects in each parent component, but the dead code is misleading and should be migrated to `useAccountEffect` or removed.
- **marketplace `requireChainId()` not-connected branches are dead code** (Tanya's LOW). `getChainId()` defaults to mainnet (`return chainIdFromSessionStorage || 1`) before ever returning `undefined`/`null`. Net effect: reads on first page load (before chain is set in sessionStorage) silently hit mainnet, even if the URL says e.g. `/gnosis/...`. UX risk: wrong-chain reads displayed as if they were the user's selected chain.
- **marketplace ListServices `mint.jsx`/`update.jsx` pass raw strings for `uint32[]` and `uint32` args**. ListAgents and ListComponents use `.map(BigInt)`. Viem coerces strings → bigints internally so it works, but the inconsistency is a footgun for the next reader.
- **`build` `getOwnerIncentives` destructure as tuple** (`apps/build/components/DevIncentives/requests.ts:103-108`) — works because viem's `Result` type is both array- *and* object-accessible, but the consumer at `IncentivesForThisEpoch.tsx:118-119` re-accesses by name. Either fully commit to one style or the other.
- **Root `web3: \"1.10.4\"` is now dead weight** in `package.json` (verified: no source imports). Removing it would shrink `node_modules` and close a supply-chain surface. Out of scope for this PR but worth a cleanup ticket.
- **CSP follow-ups** the PR body already calls out (`pulse.walletconnect.org`, `api.web3modal.org/appkit/v1/config`) — confirmed missing in `libs/common-middleware/src/lib/cspHeader.ts`. Non-blocking; apps function with console noise.
- **`Web3ModalProvider.tsx` filename rename** still pending (PR body acknowledges). Worth doing now while the surface is fresh in everyone's head.

---

### What's done well

- Bigint handling is careful: `getTotalForAllServices` casts to `Number` at the boundary with an explanatory comment (`apps/marketplace/components/ListServices/utils.tsx:72-79`); `getUnitPoint` wraps uint8 fields with `BigInt()` before mixing with uint96 (`apps/build/components/DevIncentives/requests.ts:267`).
- Centralized `params.ts` per app cleanly contains the `as Abi` casts with a clear TODO explaining the path to remove them.
- govern transports-fallback bugfix is correct (`http(RPC_URLS[id] || chain.rpcUrls.default.http[0])` — the prior `http(RPC_URLS[id]) || chain.rpcUrls.default.http[0]` could never fall through).
- The Binance custom-wallet integration via `connectorsForWallets` (`apps/contribute/components/Login/config.ts`) is clean.
- `qr@0.5.5` resolution-pin to unblock RainbowKit WC QR is a real fix, well-explained in the commit.
- All 7 apps' `CLAUDE.md` updated to reflect RainbowKit (good discipline for a migration of this size).

---

### Coverage I did NOT exercise

- **Runtime testing** against live Vercel previews. Would specifically retest: bond write paths, operate write paths, marketplace step-3 multisig flow on a real Safe — same gaps the PR body already flags as deferred.
- **`yarn.lock`** content (3,400+ lines) — CI gates already cover integrity/origin.
- **Solana / SVM flows** — code review only, can't exercise without testnet SOL.

Tanya-atatakai (dismissed):
## Re-review — all prior must-fix items resolved ✅

Re-checked against head `aedbfe8e`. Verifications below.

### Fixes verified

| Item | Where | Status |
|---|---|---|
| HIGH `number * bigint` crashes (my prior review) | `ca3af6ee` — both sites use `0n` + `BigInt()` wrap | ✅ |
| `isHashApproved` polling leak | `helpers.jsx` — 5-min timeout, `AbortSignal`, single `cleanup()` called from every exit path | ✅ |
| `getContractEvents` `fromBlock: 0n` rejected by RPC | `utils.jsx:170` — bounded to `startingBlock - 9_000` (under Alchemy 10k window) | ✅ |
| Contribute multicall blowing gas / payload budget | `nftDetails.ts` — chunked at 100 `ownerOf` calls per batch | ✅ |
| SSR persistence half-wired | All 7 apps now configure `cookieStorage` and pass `cookieToInitialState(wagmiConfig)` as `initialState` | ✅ |
| Safe-WC redirect monkey-patch | Lifted to shared `useSuppressSafeWcRedirect` hook in `libs/util-functions` — `startsWith` filter + ref-checked cleanup; applied to both `launch` and `govern` | ✅ |

### Minor: PR body is slightly stale

The body still claims "*`isMounted` defer is now applied uniformly across all 7 apps*", but commits `7ae0b53d` and `fd9fa477` actually *removed* the gate from six apps — only `contribute` keeps it (with an inline comment that it's for `ServiceStatus` / sidebar hydration, unrelated to the WC race). Functionally this is fine: SSR-via-cookies + the `qr@0.5.5` resolution pin cover the WC "invalid border=0" race that motivated the defer. Worth a quick body edit so future readers aren't confused.

### Non-blocking follow-ups

The PR body's "Known follow-ups" section already enumerates the remaining nits I'd otherwise call out (`useAccount({onConnect})` → `useAccountEffect`, marketplace `requireChainId` dead branches, raw-string vs `BigInt()` uint args, dropping root `web3@1.10.4`, restoring the 6 `describe.skip`'d marketplace suites, CSP additions for `pulse.walletconnect.org` + `api.web3modal.org`, `Web3ModalProvider` rename). None block merge.

LGTM 🚀

Tanya-atatakai:
## 🔴 Regression — gas headroom lost across all migrated write paths

> **Update (2026-05-28):** rewriting again — verified fix below now lands a successful `voteForNomineeWeightsBatch` on Tenderly. Key correction: viem's `simulateContract` **does not** return `gas` on the `request` object (this is documented — it runs `eth_call`, not `eth_estimateGas`). Any post-migration code that tries `gas: request.gas ? ... : undefined` is a no-op. The fix has to call `estimateContractGas` explicitly.

A live test on a Tenderly mainnet fork hit **`out of gas` at `gasUsed === gasLimit` (507,839 / 507,839)** on `voteForNomineeWeightsBatch`. Verified on the executed-tx trace; not a Tenderly artifact.

### What changed

Pre-migration write paths fell into two patterns, both of which gave the broadcast tx headroom:

**Pattern A — explicit helper buffer (`gasLimit: estimatedGas`):**
```js
const estimatedGas = await getEstimatedGasLimit(createLockFn, account); // Math.ceil(estimate * 1.2)
const fn = createLockFn.send({ from: account, gasLimit: estimatedGas });
```
`getEstimatedGasLimit` applied `* 1.2`, and `gasLimit` is a recognized web3.js `.send` option — the buffer reached the wallet.

**Pattern B — wallet's implicit buffer (`{ from, estimatedGas }` typo):**
```js
const estimatedGas = await getEstimatedGasLimit(voteFn, account);
const fn = voteFn.send({ from: account, estimatedGas }); // ← not a valid web3.js option, silently dropped
```
The helper's value was discarded. The tx went out with no explicit `gas`, so the wallet (MetaMask, Rabby) ran its own `eth_estimateGas` and added its own ~25% internal buffer before submission.

**Post-migration**, every migrated write now does:
```ts
const { request } = await simulateContract(wagmiConfig, {...});
const hash = await writeContract(wagmiConfig, request);
```
- `simulateContract` runs `eth_call` only — `request.gas` is `undefined` ([viem docs](https://viem.sh/docs/contract/simulateContract)).
- `writeContract` then calls `eth_estimateGas` internally and uses the raw result as the explicit `gas` in the broadcast tx.
- Because `gas` is now explicit on the tx, wallets stop second-guessing and use it verbatim.

Both Pattern A and Pattern B paths lose their headroom: helper buffer gone, wallet buffer suppressed.

### All 25 migrated callsites that lost headroom

| Layer | App | Callsite | Pre-migration headroom source |
|---|---|---|---|
| 🔴 Vote paths (Pattern B — wallet buffer) | govern | `voteForNomineeWeightsBatch`, `castVote`, `revokeRemovedNomineeVotingPower` | Wallet ~25% |
| 🟠 Lifecycle paths (Pattern A — helper 1.2×) | govern | `createLock`, `increaseAmount`, `increaseUnlockTime`, `withdraw`, `checkpoint`, `depositServiceDonationsETH` | Helper 20% |
| 🟠 Lifecycle paths (Pattern A) | bond | `BondingProducts/Bonding/Deposit/useDeposit.jsx` — `fnApprove` | Helper 20% |
| 🟠 Lifecycle paths (Pattern A) | contribute | `Leaderboard/MintNft/utils.ts` — `mintFn`<br>`Profile/Staking/requests.ts` — `restakeFn`<br>`Staking/requests.ts` — stake flow<br>`VotingPower/requests.ts` — `delegateFn` | Helper 20% |
| 🟠 Lifecycle paths (Pattern A) | launch | `common-util/functions/web3.ts` — generic `methodFn` wrapper (every launch write) | Helper 20% |
| 🟠 Lifecycle paths (Pattern A) | marketplace | `common-util/functions/requests.ts` — `claimFn`<br>`ListAgents/mint.jsx` — `createFn`<br>`ListComponents/mint.jsx` — `createFn`<br>`ListServices/mint.jsx` — `createFn`<br>`ListServices/update.jsx` — `updateFn`<br>`ListServices/ServiceState/utils.jsx` — `terminate`, `activateRegistration`, `registerAgents`, `deploy`, `unbond` | Helper 20% |

### Verified fix (works on Tenderly)

Because `simulateContract` doesn't return gas, you must call `estimateContractGas` separately. Tested pattern that landed a successful batch vote:

```ts
import { getPublicClient, simulateContract, writeContract, waitForTransactionReceipt } from '@wagmi/core';

const { request } = await simulateContract(wagmiConfig, {
  ...voteWeightingParams,
  functionName: 'voteForNomineeWeightsBatch',
  args: [nominees, chainIds, weights],
  account,
});

const publicClient = getPublicClient(wagmiConfig, { chainId: MAINNET_CHAIN_ID });
const gasEstimate = await publicClient.estimateContractGas({
  ...voteWeightingParams,
  functionName: 'voteForNomineeWeightsBatch',
  args: [nominees, chainIds, weights],
  account,
});

const hash = await writeContract(wagmiConfig, {
  ...request,
  gas: (gasEstimate * 120n) / 100n, // 20% headroom
});
return waitForTransactionReceipt(wagmiConfig, { hash });
```

Net cost: one extra RPC round-trip on the write path (`eth_call` for simulation + `eth_estimateGas` for the buffer). Acceptable since writes are already interactive.

**Centralized helper (recommended)** — drop into `libs/util-functions` so the multiplier lives in one place:
```ts
import type { Config } from '@wagmi/core';
import { getPublicClient } from '@wagmi/core';
import type { EstimateContractGasParameters } from 'viem';

export const estimateGasWithBuffer = async (
  config: Config,
  call: EstimateContractGasParameters & { chainId?: number },
  bps = 120n,
): Promise<bigint> => {
  const publicClient = getPublicClient(config, { chainId: call.chainId });
  const estimate = await publicClient.estimateContractGas(call);
  return (estimate * bps) / 100n;
};

// usage at every migrated callsite:
const { request } = await simulateContract(wagmiConfig, params);
const gas = await estimateGasWithBuffer(wagmiConfig, params);
const hash = await writeContract(wagmiConfig, { ...request, gas });
```

**All 25 callsites above need this treatment** — anywhere `getEstimatedGasLimit` was previously used. Quick patterns to check by app:

- `apps/bond/components/BondingProducts/Bonding/Deposit/useDeposit.jsx`
- `apps/contribute/components/Leaderboard/MintNft/utils.ts`, `Profile/Staking/requests.ts`, `Staking/requests.ts`, `VotingPower/requests.ts`
- `apps/govern/common-util/functions/requests.ts` (9 writes in this file)
- `apps/launch/common-util/functions/web3.ts` (single wrapper covers every launch write)
- `apps/marketplace/common-util/functions/requests.ts`, `components/ListAgents/mint.jsx`, `ListComponents/mint.jsx`, `ListServices/{mint,update}.jsx`, `ListServices/ServiceState/utils.jsx`

Why wallets won't paper this over: most wallets only auto-buffer when `gas` is unset on the request. viem's `writeContract` always sets it (via its internal `eth_estimateGas` call). So this regression bites every path with non-deterministic gas — `VoteWeighting`/veOLAS checkpoint loops, conditional storage writes, multi-iteration registration/approval flows, arbitrum-bridge writes in govern.

### Verdict

Should land before merge. Verified the proposed pattern locally — `voteForNomineeWeightsBatch` lands successfully on a Tenderly fork with `gasUsed` ≈ 540k against a `gasLimit` ≈ 609k after the buffer.

rajat2502 (approved):
Ship it. The high-severity bugs (polling leak, fromBlock: 0 RPC rejection, multicall gas overflow) were caught and fixed within the PR. Supply-chain hygiene is solid — exact pins, qr@0.5.5 rationale documented, audit gates clean. Remaining items are either tracked follow-ups or pre-existing conditions, none introduced by this PR.

Tanya-atatakai (approved):

mohandast52:
## Re-review (HEAD: `d392d418c`)

Re-walked the diff after the 23 commits since my last comment, verified each fix against the current code, and rebuilt marketplace + launch + contribute locally. CI is green across all 7 Vercel previews and every supply-chain gate.

### Resolved from the previous review

| # | Original finding | Fix | Status |
|---|---|---|---|
| 1 | `isHashApproved` no timeout / no cleanup | `ffe2c76aa` — 5 min `MAX_POLL_MS`, optional `AbortSignal`, `cleanup()` consolidates clearInterval + clearTimeout + listener removal across all exit paths | ✅ |
| 2 | `getContractEvents` with `fromBlock: 0n` | `ffe2c76aa` — bounded to `startingBlock - 9_000n` (under Alchemy's 10k cap, ~30h on mainnet) | ✅ |
| 3 | Contribute single-shot multicall over thousands of mints | `354988865` — chunked at 100 calls per multicall, sequential batches | ✅ |
| 4 | SSR persistence half-configured across 5 apps | `d5b9fe4bd` — `cookieStorage` + `cookieToInitialState(wagmiConfig)` wired in every app's `_app` and provider | ✅ |
| 5 | Provider-mount `isMounted` defer inconsistent | `d5b9fe4bd` → `7ae0b53d7` → `fd9fa4773` — `isMounted` gate dropped from every app once the WC race no longer reproduces; CLAUDE.md notes added to launch/operate explaining the SSR rationale ("pages calling `useConfig` at top-render fail SSR/static export when the provider isn't there") | ✅ |
| 6 | `launch` `window.open` patch fragile + scope-limited | `d5b9fe4bd` — lifted to shared `libs/util-functions/useSuppressSafeWcRedirect`. Tighter `startsWith('https://app.safe.global')` match, cleanup gated on `window.open === patched` so it can't un-wrap a sibling library's patch, and now applied to govern too (the other Safe-via-WC path) | ✅ |
| — | CSP follow-up (`pulse.walletconnect.org`, `api.web3modal.org`) | `d392d418c` — both added to `ALLOWED_ORIGINS`. Real user-visible regression caught (missing mobile-wallet deeplink list when CSP blocks `api.web3modal.org/appkit/v1/config`) | ✅ |
| — | Implicit `http(undefined)` RPC fallback | `d5b9fe4bd` — explicit `http(RPC_URLS[chain.id] \|\| chain.rpcUrls.default.http[0])` now uniform across bond/build/launch/operate, matching govern/marketplace/contribute | ✅ |

### New finding caught & fixed since last review (critical)

**Gas headroom silently dropped on every write path** — `0dd17b11e` (caught by another reviewer, not me; I missed it last time and would've shipped a broken rollup).

Root cause is subtle: viem's `simulateContract` runs `eth_call` only and never sets `gas` on the request. wagmi's `writeContract` then calls `eth_estimateGas` itself and uses the raw lower-bound estimate as an *explicit* `gas` field on the broadcast tx. Wallets only apply their own ~25% safety buffer when `gas` is unset — with viem's explicit value, MetaMask uses it verbatim. Verified on Tenderly: `voteForNomineeWeightsBatch` hit `gasUsed === gasLimit (507839/507839)`. Pre-migration web3.js code passed `gasLimit: estimate * 1.2` via a `getEstimatedGasLimit` helper that didn't survive the port.

Fix introduces `libs/util-functions/estimateGasWithBuffer` and wires it into 25 callsites across 6 apps (govern×10, marketplace×11, contribute×4, launch×2, bond×2). Pattern is:

```js
const callParams = { ...params, functionName, args, account };
const { request } = await simulateContract(wagmiConfig, callParams);
const gas = await estimateGasWithBuffer(wagmiConfig, callParams);
const hash = await writeContract(wagmiConfig, { ...request, gas });
```

The follow-up `cd99262b9` guards against `getPublicClient` returning undefined with a meaningful error. Good defence-in-depth.

### Other meaningful new fixes

- **`d96e93b4b`** — `tmp` 0.2.4 → 0.2.6 clears GHSA-ph9p-34f9-6g65 in the supply chain.
- **`f85a0a160`** — `SwitchNetworkButton` shows "Confirm in wallet" while `isPending`, giving users a visible cue that focus moved off-app. Mirrors RainbowKit's chain-modal indicator.
- **`b8a6e80f6`** / **`3107349b4`** — Nx + Next config plumbing fixes so `nx serve` works on the webpack-build apps and `withNx` doesn't collide with `outputPath`. Infrastructure, no runtime impact.
- **`031dc80ae`** / **`f046bbc7d`** — TS overload inference fixes for the new gas helper and barrel-import order.
- **`32877ba05`** — Tenderly virtual-testnet env-var path now documented in govern's CLAUDE.md, which is presumably where the gas-headroom bug was reproduced.

### Still open (none blocking, repeating from prior review)

- 8 marketplace test suites still `describe.skip`'d with `TODO(viem-migration)`. The bigint-bond accumulator paths fixed in `ca3af6ee5` remain untested by automated regression coverage — worth a follow-up ticket so the next ABI change there gets caught.
- Dead wagmi v1 `useAccount({ onConnect, onDisconnect })` callbacks still in bond/build/marketplace/govern LoginV2. Functionally compensated by parallel useEffects, but should be migrated to `useAccountEffect` (or just deleted).
- Marketplace `requireChainId()` not-connected branches at `apps/marketplace/common-util/functions/index.ts:54-61` are still dead code (Tanya's LOW).
- Marketplace ListServices `mint.jsx`/`update.jsx` still pass raw strings for `uint32[]` / `uint32` args while ListAgents/ListComponents use `.map(BigInt)` — viem coerces internally so it works, but the inconsistency is a footgun.
- Marketplace `build` `getOwnerIncentives` destructure mixes tuple-style and named-style access (works thanks to viem's `Result` shape, but inconsistent).
- Root `web3: "1.10.4"` still in `package.json` despite no remaining imports — dead-weight dep, removing would shrink `node_modules` and close a supply-chain surface.
- `Web3ModalProvider.tsx` filename rename to `WalletProvider.tsx` still pending across apps.

### Verdict

**Mergeable.** Both `Tanya-atatakai`'s blocking HIGHs and all three of my must-fix items are resolved with correct, well-explained patches. The independently-caught gas-headroom regression closes what would've been the worst silent failure mode of this rollup. Open items are cleanup, not correctness — happy to see them shipped as follow-ups.

mohandast52 (approved):
Approving on the basis of the re-review in https://github.com/valory-xyz/autonolas-frontend-mono/pull/416#issuecomment-4567520588 — all blocking items from this round and the prior round are resolved; remaining open items are cleanup, not correctness.

PR #418, opened 2026-06-05
chore(deps): remove unused web3 1.x + bump axios to 1.17.0

mohandast52:
Two related dependency-hygiene changes. (The axios bump was folded in to unblock CI — it is a repo-wide failure, not caused by the web3 removal; see below.)

## 1. Remove unused `web3` (web3.js 1.x)
- **Zero source imports** anywhere (apps/libs + build configs); `yarn why web3` shows it's required only by the root manifest.
- LGPL-3.0, drags in 20 `web3-*` sub-packages (also LGPL-3.0) + a large transitive tree (`es5-ext`, `secp256k1`, old `@ethereumjs` copies). Removing it deletes **~1200 lines of yarn.lock**.
- Prunes 2 now-stale install-hook allowlist entries (`es5-ext`, `secp256k1`) that only web3 pulled.
- Clears 21 entries from the license-allowlist gate (#417) once both land.

## 2. Bump `axios` 1.15.2 → 1.17.0
- `axios <1.16.0` has multiple **newly-published high-severity advisories** (prototype-pollution MitM via `config.proxy`, ReDoS via cookie-name injection, proxy-auth credential leaks on redirect).
- Bumped both the direct dependency **and the resolution** (forces transitive copies to the fixed version).
- **This failure is pre-existing on `main`** — `audit:prod` fails there identically. Not introduced by this branch; folded in to get CI green.

## Verification
- ✅ `yarn audit:prod` passes (0 high/critical).
- ✅ **All 9 apps + 3 libs build** with `--skip-nx-cache` (twice — after web3 removal, and after the axios bump).
- ✅ install-hooks audit passes.
- ✅ Failing unit tests (`ui-components`/`util-functions` JSX-transform, `marketplace` timeouts) fail **identically on `main`** — pre-existing local test-env issues, unrelated to these changes.

## Follow-up
Once this and #417 land, delete the 21 temporary `web3`/`web3-*` exemptions from `.supply-chain/license-allowlist.json`.

PR #419, opened 2026-06-05
chore(marketplace): replace @gnosis.pm/safe-contracts (GPL-3.0) with local helpers

mohandast52:
Removes the last **GPL-3.0 code shipped in a bundle**: `@gnosis.pm/safe-contracts`, whose `dist/` JS was `require()`d at runtime by the marketplace multisig service-registration flow.

## Change
- New GPL-free module [`apps/marketplace/common-util/functions/safeTransactions.js`](apps/marketplace/common-util/functions/safeTransactions.js) implementing the three helpers used (`buildSafeTransaction`, `encodeMultiSend`, `buildMultiSendSafeTx`) per the on-chain MultiSend.sol layout.
- [`3rdStepFinishedRegistration/utils.jsx`](apps/marketplace/components/ListServices/ServiceState/3rdStepFinishedRegistration/utils.jsx) now imports the local module instead of `require('@gnosis.pm/safe-contracts')`.
- Dropped the package + its now-inert `@gnosis.pm/safe-contracts/ethers` resolution.

## Why it is safe — proven, not asserted
These are **pure functions**, so identical output ⇒ identical behaviour for every downstream `getTransactionHash` / `execTransaction` / EIP-712 signing call (none of which changed).

- **Differential fuzz vs the real `@gnosis.pm/safe-contracts@1.3.0`: 60,256 cases, 0 mismatches.** Covers all 256 `buildSafeTransaction` default-combinations + 20,000 randomized multi-tx batches (1–8 txs, data 0–200 bytes incl. empty/odd lengths, varying operation/value/nonce), asserting `deepStrictEqual` on `buildSafeTransaction`, `encodeMultiSend`, and `buildMultiSendSafeTx`.
- **Committed golden-value test** ([safeTransactions.test.js](apps/marketplace/tests/common-util/functions/safeTransactions.test.js)) locks the exact bytes captured from the original — no dependency on the GPL package.
- **All 9 apps + 3 libs build** with `--skip-nx-cache`.

## Residual (belt-and-suspenders)
The encoding is provably identical, but this is the live Safe-multisig deploy path — a testnet run of the documented two-account flow (top of `utils.jsx`) before merge is recommended as end-to-end confirmation of the *unchanged* surrounding logic.

## Sequencing
Branched from `main`, so `audit:prod` will be red until #418 (the axios fix) lands on `main` and this rebases. Once #417 + #418 + this all land, the `@gnosis.pm/safe-contracts` exemption can also be deleted from the license allowlist.

### DIFF
--- a/apps/marketplace/common-util/functions/safeTransactions.js
+++ b/apps/marketplace/common-util/functions/safeTransactions.js
@@ -1,0 +1,60 @@
+import { ethers } from 'ethers-v5';
+
+/**
+ * GPL-free reimplementations of the three `@gnosis.pm/safe-contracts` helpers the
+ * marketplace multisig flow used (`buildSafeTransaction`, `encodeMultiSend`,
+ * `buildMultiSendSafeTx`). They implement the on-chain Safe MultiSend encoding
+ * (defined by `MultiSend.sol`: operation ++ to ++ value ++ dataLength ++ data),
+ * so the bytes are identical to the originals — proven byte-for-byte against the
+ * real package in `tests/common-util/functions/safeTransactions.test.js`.
+ *
+ * Replaced `@gnosis.pm/safe-contracts` (GPL-3.0), whose `dist/` JS shipped into
+ * the marketplace bundle. Uses `ethers-v5` to match the version safe-contracts
+ * used internally and the version this multisig flow already runs on.
+ */
+
+const { AddressZero } = ethers.constants;
+
+/** Fill a Safe transaction template with the Safe contract's zero-defaults. */
+export const buildSafeTransaction = (template) => ({
+  to: template.to,
+  value: template.value || 0,
+  data: template.data || '0x',
+  operation: template.operation || 0,
+  safeTxGas: template.safeTxGas || 0,
+  baseGas: template.baseGas || 0,
+  gasPrice: template.gasPrice || 0,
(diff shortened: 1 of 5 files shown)


rajat2502 (approved):
LGTM — approving.

Removes the last GPL-3.0 bundled dep (`@gnosis.pm/safe-contracts`) by reimplementing the three MultiSend helpers locally, with equivalence proven by a golden test + characterization test against the real package's captured bytes. Ran the new test suite in a worktree: 5/5 pass.

Verified:
- `onMultisigSubmit` wiring preserved 1:1; only caller (`index.jsx:361`) unaffected — signature unchanged, new exports additive.
- No other source references `safe-contracts`.
- Removing the `ethers@5.7.2` lock entry does **not** affect the `ethers-v5` alias — that's a separate lock entry (`ethers-v5@npm:ethers@5.7.2`), untouched.

Non-blocking:
- Recommend the documented two-account testnet run of the live Safe-multisig deploy path before merge (as already noted in the PR) — the helpers are byte-identical, but the surrounding sign/exec wiring isn't unit-tested.
- Per the PR's sequencing note: don't forget to delete the `@gnosis.pm/safe-contracts` license-allowlist exemption in the cleanup PR once #417/#418 land and this rebases.

Tanya-atatakai (approved):

PR #423, opened 2026-06-17
fix: [BDMARK-2192] resolve duplicate meta titles & descriptions (marketplace, govern, build, bond)

atepem:
## Summary

Fixes the duplicate **meta title** and **meta description** issues reported by SEMrush across the Marketplace, Govern, Build, and Bond apps.

Investigation showed the flags fell into three buckets — only the latter two were live code bugs; the first was stale crawl data already addressed by the earlier "add per-page metadata" commits.

### 1. Missing canonical tags → query-param duplicates
URLs like `?proposalId=…`, `?legacy=true`, and `?activity=true` are variants of the same page but were indexed as separate URLs with identical titles. The `Meta` component in **all four apps** now emits `<link rel="canonical" href={cleanUrl} />` built from the query-string-free `pageUrl`/`path`, consolidating these variants onto one canonical URL. Also hardened the govern/build URL builder against an `undefined` path.

Covers: `govern.olas.network/proposals?proposalId=…`, `marketplace…/gnosis/ai-agents?legacy=true`, `marketplace…/gnosis/ai-agents/2235?activity=true`.

### 2. Build `/paths/[id]` meta not server-rendered
The 6 `/paths/*` pages all served the **default** title/description in their SSR HTML. `pathData` was loaded in a `useEffect` (never runs on the server) and `<Meta>` was gated behind the loading spinner, so the per-path meta never reached the initial HTML.

- `useFetchPathData` now resolves `pathData` **synchronously** via `useMemo` (the route param is available server-side because `_app` uses `getInitialProps`, disabling Automatic Static Optimization). Only the markdown body remains async.
- `paths/[id].tsx` renders `<Meta>` as soon as `pathData` is known, including the loading state.

Covers all 6 build path pages (ecosystem-opportunities, general-mechs-tool, shorts-mech-tool, prediction-agents-trading-strategy, build-your-own-service, prediction-agents-mechs-ai-tool) — duplicate titles **and** descriptions.

### 3. Cross-network title collisions (marketplace)
The same list type on different chains rendered identical titles (`/ethereum/ai-agents` and `/gnosis/ai-agents` were both "AI Agents"), and every `/[network]` dashboard rendered the bare brand title. Added `getNetworkDisplayName()` (`common-util/functions`) to map the route slug (e.g. `arbitrum-one`) to a display name (e.g. `Arbitrum One`), used in the AI Agents list/mint pages and the network dashboard to produce titles like `AI Agents on Gnosis` / `Ethereum Registry`.

## Verification
- Built the Build app and inspected raw `curl` output (what a non-JS crawler sees):
  - **Before:** every `/paths/*` page → `<title>Build | Olas</title>` + default description.
  - **After:** `Ecosystem Opportunities | Build | Olas`, `General Mechs Tool | Build | Olas`, etc., each with its own description and a per-path canonical — all present in the initial server HTML.
- `tsc --noEmit` clean on build, marketplace, govern, bond.
- `nx lint` clean (0 errors) on marketplace, govern, bond. (The `build` lint target throws a pre-existing "Nothing to lint" config error, identical on `main` — unrelated to this change.)

## Notes
- Several SEMrush entries (e.g. marketplace `/ethereum/components`, `/ethereum/components/mint`, `/ethereum/agent-blueprints`) already render distinct titles in SSR HTML thanks to the earlier per-page-metadata work — those were stale crawl results and clear on re-crawl with no code change.
- These are crawler-facing fixes; SEMrush will reflect them after deploy + re-crawl.
- App-level `CLAUDE.md` files updated to document the canonical/SSR-meta conventions so they aren't regressed.


apps/govern/components/Meta.tsx
```diff
@@ -15,12 +15,14 @@ type MetaProps = {
...
   const title = pageTitle ? `${pageTitle} | ${SITE_TITLE}` : SITE_TITLE;
-  const url = `${SITE_URL}/${pageUrl}`;
+  const url = `${SITE_URL}/${pageUrl || ''}`;
 
   return (
     <Head>
       {/* <!-- Primary Meta Tags --> */}
       <title>{title}</title>
+      {/* Canonical URL (without query string) to consolidate duplicate URL variants */}
+      <link rel="canonical" href={url} key="canonical" />
```
mohandast52, line 25: Adding `rel=canonical` to the shared `Meta` makes it authoritative for **every** caller. Two existing govern call sites that were fine when `pageUrl` only fed `og:url` now emit incorrect canonicals:

**1. `pages/veolas.tsx:9` passes `pageUrl="veOlas"`** → canonical `https://govern.olas.network/veOlas`, but the route is `/veolas` (file `pages/veolas.tsx`, nav link `Balance.tsx` uses `href="/veolas"`). Next pages-router paths are case-sensitive, so `/veOlas` doesn't serve this page — the canonical points at a non-equivalent (404) URL and is self-defeating. Fix: `pageUrl="veolas"`.

**2. Contract detail pages** — `pages/contracts/[address].tsx:9` passes `pageUrl="contracts"`, and the inner `<Meta>` at `components/Contract/index.tsx:141` (rendered once the contract loads) passes **no** `pageUrl`. Since both share `key="canonical"`, the later-rendered one wins:
- SSR (inner `<Meta>` not yet rendered, page shows skeleton) → canonical `/contracts` (every contract page collapses onto the list).
- Client, after the contract loads → inner `<Meta>` wins → canonical `https://govern.olas.network/` (the **homepage**).

So a JS-rendering crawler ends up seeing the homepage as the canonical for every contract detail page. Give the contract detail a stable self-referential canonical — pass `pageUrl={`contracts/${formattedAddress}`}` to the inner `<Meta>` (and ideally make the page-level one match) so SSR and CSR agree.

(Marketplace/bond/build callers are all self-referential, so they're unaffected — this is govern-specific.)

apps/build/pages/paths/[id].tsx
```diff
@@ -31,21 +31,32 @@ const markdownOptions = {
...
-        </div>
+        <Typography.Title level={2}>Path not found</Typography.Title>
       </Container>
     );
   }
 
-  if (!pathData) {
+  // Render Meta as soon as the path entry is resolved (synchronously, incl. SSR)
+  // so crawlers always receive the per-path title/description.
+  if (loading || !pathData) {
```
Comment (mohandast52) [inline apps/build/pages/paths/[id].tsx:51]: Optional nit: after the `if (!loading && !pathData)` guard above, the only remaining way for `!pathData` to be true is when `loading` is also true — so this is equivalent to `if (loading)`. The `|| !pathData` is harmless (the `{pathData && <Meta/>}` below already guards the null case), just slightly redundant. Fine to leave.

apps/marketplace/common-util/functions/index.ts
```diff
@@ -19,13 +19,29 @@ import { RpcUrl } from 'libs/util-functions/src/lib/sendTransaction/types';
...
 export const getModalProvider = () => (window as any)?.MODAL_PROVIDER;
 
+/**
+ * Maps a route network slug (e.g. "ethereum", "arbitrum-one") to its display
+ * name (e.g. "Ethereum", "Arbitrum One"). Used to differentiate page meta titles
+ * across networks so the same list type on different chains is not a duplicate.
+ */
+export const getNetworkDisplayName = (network?: string | string[]): string | null => {
+  if (typeof network !== 'string') return null;
+  const chain = ALL_SUPPORTED_CHAINS.find((c) => c.networkName === network);
```
mohandast52: Optional: this match is case-sensitive, while route validation in `useHandleRoute.jsx` is case-insensitive (`toLower(e.networkName) === toLower(name)`). A mixed-case slug like `/Gnosis/ai-agents` would still render (routing accepts it) but `getNetworkDisplayName` returns `null`, dropping the network from the title and leaving a mixed-case self-canonical. Edge case (internal links always use the lowercase slug) — lowercasing `network` here would harden it. Not blocking.

mohandast52 (changes requested):
## Review — duplicate meta titles & descriptions

Reviewed the full diff plus the surrounding call sites across all four apps. The framing into three buckets is right and the overall approach is sound.

### ✅ Verified good
- **Canonical tags** are added with `key="canonical"`, so the page-level canonical correctly dedupes against the default `<Meta />` rendered in each app's `_app`. (Without the key there'd be two conflicting canonicals.)
- **`getNetworkDisplayName`** maps the route slug correctly. The `[network]` slug is `networkName` = `kebabCase(chain.name)` (confirmed via `store/setup.ts` `setChainId`/`setVmInfo` and `common-util/hooks/useHandleRoute.jsx`), and the lookup keys on `networkName` — so `arbitrum-one → Arbitrum One`, `op-mainnet → OP Mainnet`, `mode-mainnet → Mode Mainnet` etc. all resolve, with a graceful `null` fallback for unknown slugs.
- **Build SSR meta** via `useMemo`: `pathData` now resolves synchronously, so `<Meta>` lands in the initial server HTML. `_app` uses `getInitialProps` (ASO disabled) → `router.query.id` is populated server-side, so the mechanism checks out. The loading / not-found / loaded branches in `paths/[id].tsx` are correct and the initial client render matches SSR (`loading=true` + resolved `pathData`), so no hydration mismatch.
- **`|| ''` hardening** in build/govern `Meta` is meaningful — it fixes the default `<Meta />` (no `path`/`pageUrl`), which previously produced a `.../undefined` URL.
- Marketplace, bond, and build `Meta` callers are all self-referential, so their canonicals are correct. No component-level `<Meta>` outside govern.

### 🔧 Please address before merge (govern only)
Making the shared `Meta` emit an *authoritative* canonical surfaces two pre-existing `pageUrl` values that were harmless when they only fed `og:url`. Details inline on `apps/govern/components/Meta.tsx`:
1. **`pages/veolas.tsx`** → canonical resolves to `/veOlas`, but the route is `/veolas` (case mismatch → self-defeating canonical).
2. **Contract detail pages** → canonical resolves to the **homepage** client-side (the inner `<Meta>` in `components/Contract/index.tsx` passes no `pageUrl`) and to `/contracts` in SSR — neither is self-referential.

Both ship to a live app, so worth correcting (or confirming the contract→`/contracts` consolidation is intentional — the homepage canonical is wrong either way).

### Nits (optional, non-blocking)
A redundant condition in `paths/[id].tsx` and a case-sensitivity edge case in `getNetworkDisplayName` — inline.

Holding approval on the two govern canonical fixes; everything else looks good. (Did not re-run `tsc`/lint myself — relying on the PR's reported clean runs; the changes are type-compatible on inspection.)

Tanya-atatakai (commented):
## Code Review

Clean, well-scoped fix addressing three distinct root causes (canonical tags, SSR-able per-path meta in build, network-qualified marketplace titles). I verified the key correctness assumptions:

- ✅ `ALL_SUPPORTED_CHAINS` / `networkName` / `networkDisplayName` exist in `common-util/Login/config`; `getNetworkDisplayName` is null-safe (handles `undefined` / `string[]`).
- ✅ build `SITE_URL` has a trailing slash, so `path="paths/x"` yields a correct `…/paths/x` canonical despite the path lacking a leading `/`.
- ✅ All canonical/og/twitter tags use `key=…`, so Next dedupes across re-renders.
- ✅ `useMemo` resolution makes `pathData` available at SSR, which is what lands the meta in initial HTML.

### Minor (non-blocking)
- `apps/build/pages/paths/[id].tsx` — the second guard `if (loading || !pathData)` has an unreachable `!pathData` branch (the first guard already returns for `!loading && !pathData`), so `{pathData && <Meta/>}` is effectively always truthy there. Could simplify to `if (loading)`. Behavior is correct.
- The `<Meta …/>` block is duplicated verbatim in the loading and final branches — could hoist to a local const.
- Pagination canonical: list pages use `?page=N`; canonical strips it and consolidates to the page-1 URL. Conventional and likely intended — flagging as a conscious SEO decision.
- No tests added. A small unit test for `getNetworkDisplayName` + a Meta snapshot asserting the canonical tag would lock this in cheaply.

### Follow-up: same issues exist in untouched apps
The PR matches the SEMrush report scope, but two other live apps have the **same duplicate-title bug classes** and will likely be flagged on the next crawl:

- **launch** (same patterns, not fixed):
  - `pages/[network]/my-staking-contracts/*` share `pageTitle="My Staking Contracts"` / `"Staking Contract Details"` across networks → cross-network duplicate titles (same as marketplace fix here).
  - `pages/paths/[id].tsx` hardcodes `pageTitle="Path Details"` for every path → duplicate titles (same class as the build fix, arguably worse as it's a literal constant).
  - `pages/nominate-contract/[id].tsx` → shared `"Nominate Contract"`.
  - No canonical tag in `components/Meta.tsx`.
- **contribute**: `pages/profile/[id].tsx` uses `pageTitle="Profile"` for all profiles → duplicate titles; no canonical.
- operate/docs: low/no risk, though their `Meta` also lacks a canonical.

Suggest a follow-up ticket to extend the canonical + per-page-title work to **launch** (and the contribute profile page).

**Verdict: looks good to merge.** Findings above are minor/optional or out-of-scope follow-ups.

apps/govern/components/Meta.tsx
atepem: Both fixed in 8a85cb384:

1. **veolas** — `pageUrl="veOlas"` → `"veolas"`, so the canonical now matches the served lowercase route.
2. **Contract detail** — made the canonical self-referential and SSR/CSR-consistent. The page-level `<Meta>` (`pages/contracts/[address].tsx`) now emits `contracts/${address}` from `router.query.address`, and the inner `<Meta>` (`components/Contract/index.tsx`) emits `contracts/${contract.address.toLowerCase()}`. Both are lowercased — the contract lookup already matches on lowercased address, so SSR and CSR resolve to the same canonical and checksummed/lowercase URL variants consolidate. No more `/contracts` (SSR) or homepage (CSR) canonical.

Confirmed via grep that the other govern callers (contracts list, docs, donate, epoch, proposals) are already self-referential.

apps/build/pages/paths/[id].tsx
Comment (atepem) [reply apps/build/pages/paths/[id].tsx:51]: Hoisted the duplicated `<Meta>` into a single `meta` const (also flagged by @Tanya-atatakai). Kept the `|| !pathData` in the guard though — dropping it to `if (loading)` makes TS lose the null-narrowing and `pathData` becomes `possibly null` in the final return (tsc TS18047). Added a comment noting it's retained for narrowing.

apps/marketplace/common-util/functions/index.ts
atepem: Hardened — `getNetworkDisplayName` now lowercases both the input slug and `networkName` before matching, so a mixed-case slug like `/Gnosis/ai-agents` resolves the display name (and avoids the mixed-case self-canonical you described).

atepem:
Thanks both for the thorough reviews. Addressed in `8a85cb38`:

**Blocking (govern canonicals) — @mohandast52**
- `veolas.tsx`: `pageUrl` `veOlas` → `veolas` (matches the served lowercase route).
- Contract detail: canonical is now self-referential and SSR/CSR-consistent — both the page-level `<Meta>` (`router.query.address`) and the inner `<Meta>` (`contract.address`) emit `contracts/<address>` lowercased, so they agree and casing variants consolidate. No more `/contracts` (SSR) / homepage (CSR) canonical.

**Nits addressed**
- `getNetworkDisplayName`: case-insensitive slug match (handles mixed-case slugs).
- `paths/[id].tsx`: hoisted the duplicated `<Meta>` into one `meta` const. Kept `|| !pathData` in the guard — removing it makes TS lose null-narrowing in the final return (TS18047); added a comment explaining why.

**Acknowledged, intentionally not changed**
- _Pagination canonical_ (@Tanya-atatakai): stripping `?page=N` to consolidate onto page 1 is the intended SEO behaviour — same rationale as the query-param consolidation this PR is about.
- _Tests_: skipped a `getNetworkDisplayName` unit test for now because the only export path is `common-util/functions/index.ts`, which pulls the full ethers/solana/web3 surface into the test — not worth the import-weight/mock churn here. Happy to add one if you'd prefer it as a follow-up.
- _Same SSR-meta pattern in untouched apps_ (@Tanya-atatakai): real, but out of scope for this PR — worth a separate follow-up issue.

All changes verified: `tsc --noEmit` clean (build/govern/marketplace), `nx lint` clean (0 errors). Re-requesting review.

Tanya-atatakai (approved):

apps/govern/pages/contracts/[address].tsx
```diff
@@ -1,15 +1,26 @@
...
-    <ContractPage />
-  </>
-);
+const Contract = () => {
+  const router = useRouter();
+  // Self-referential canonical for the contract detail page. Lowercased so
+  // checksummed/lowercase address variants (both serve the same page) collapse
+  // onto one canonical, and so SSR matches the CSR canonical from the inner
+  // <Meta> in components/Contract (which keys on the same lowercased address).
+  const address = String(router.query.address || '').toLowerCase();
```
Comment (mohandast52) [inline apps/govern/pages/contracts/[address].tsx:12]: Non-blocking FYI: this route is auto-static-optimized (no `getInitialProps` in `_app`, no `getServerSideProps`/`getStaticProps` here), so during prerender `router.query.address` is `undefined` → `address = ''` → the canonical in the *raw* HTML is `https://govern.olas.network/contracts/` (address-less). After hydration it re-renders to the correct `contracts/<address>`, and the inner `<Meta>` in `components/Contract` (once the contract loads) also emits the same value — so JS-rendering crawlers like Googlebot get the right self-referential canonical. This fully resolves the original homepage-canonical bug; the address-less raw-HTML value is just a static-optimization residual. If you ever want it correct for non-JS crawlers, adding `getServerSideProps` to populate the param server-side would do it. No change needed for this PR.

mohandast52 (approved):
## Re-review — all review feedback addressed ✅

Thanks for the quick turnaround on commit `8a85cb3`. Both blocking items and both nits are resolved:

**Blocking (govern):**
1. **`veolas.tsx`** → `pageUrl="veolas"`. Static route, so canonical is now `/veolas` in both SSR and CSR — matches the actual route. ✅
2. **Contract detail** → both the page-level (`contracts/[address].tsx`) and the inner `<Meta>` (`components/Contract/index.tsx`) now emit a self-referential `contracts/<address>`, lowercased. Since the matcher keys on the same lowercased address, the two `key="canonical"` tags agree, so the client canonical is now correct instead of resolving to the homepage. Lowercasing also nicely collapses checksummed/lowercase address variants (both serve the same page) onto one canonical. ✅

**Nits:**
- `getNetworkDisplayName` now matches case-insensitively, aligning with `useHandleRoute`'s validation. ✅
- `paths/[id].tsx` hoists the duplicated `<Meta>` into a single `meta` const — and good catch keeping `|| !pathData` for TS null-narrowing (it's what lets `pathData` be treated as non-null in the final return); the comment makes that explicit. ✅

Verified all CI is green on the latest commit (build jobs, lint, lockfile-lint, audit, plus the govern/build/marketplace Vercel deploys).

**Non-blocking FYI (no change requested):** see the inline note — for the contract route, the *raw* prerendered HTML still emits `contracts/` (address-less) because govern doesn't use `getInitialProps`/`getServerSideProps`, so `router.query.address` isn't populated until hydration. The correct per-address canonical appears after client render, which is what JS-rendering crawlers (Googlebot) use — so this is fine for the PR's goal and strictly better than before. Just flagging in case you ever want the canonical correct for non-JS crawlers on that route too (would need server-side params).

Approving — nice work.

PR #427, opened 2026-06-23
fix(govern): stop proposals page crashing on null quorum

Tanya-atatakai:
## Problem

The Govern proposals page white-screens with:

```
TypeError: Cannot convert undefined to a BigInt
    at isQuorumReached (...)
```

## Why it happens

The crash is triggered by a **legacy Defeated proposal that has no quorum value**:

- The subgraph has **no quorum record** for it (quorum is only written when a proposal is queued/executed/cancelled — this one never was).
- We **no longer estimate its quorum on-chain**, because it was created on the **old governor contract** which has since been replaced. The current `GovernorOLAS` returns `0` for that proposal's snapshot block (its quorum-numerator checkpoint history doesn't reach back that far), so an on-chain read can't recover a meaningful value either.

We deliberately **do not special-case backfilling** this — it's legacy and the proposal is Defeated anyway, so the exact quorum figure is irrelevant. The subgraph will be updated separately in the future to record quorum for these proposals.

Until then, the UI simply must not crash when `quorum` is null. Today it calls `BigInt(quorum)` / `ethers.formatUnits(quorum)` on the null value and takes the whole list down.

## Fix — make the render path null-safe

1. `common-util/graphql/types.ts` — type `quorum` as `string | null` (matches reality; it can legitimately be absent).
2. `components/Proposals/utils.ts`:
   - `isQuorumReached` returns `false` when `quorum == null` instead of `BigInt(null)`. The proposal renders as **Defeated**, which is correct for this legacy proposal.
   - `formatWeiToEth` is null-safe, so the Quorum column renders `0` instead of throwing in `ethers.formatUnits`.
3. `hooks/useProposals.ts`:
   - Normalize the on-chain backfill fallback to `null` (not `undefined`) so the value satisfies the type.
   - Pin `useBlock({ chainId: mainnet.id })` — proposal `startBlock`/`endBlock` and `GovernorOLAS` are on mainnet, so a non-mainnet wallet otherwise yields meaningless block-number comparisons for status.

## Testing

- `yarn nx lint govern` — clean (only a pre-existing unrelated `` warning).
- `npx tsc --noEmit -p apps/govern/tsconfig.json` — passes.

image

### DIFF
--- a/apps/govern/common-util/graphql/types.ts
+++ b/apps/govern/common-util/graphql/types.ts
@@ -14,7 +14,9 @@ export type Proposal = {
   description: string;
   votesFor: string;
   votesAgainst: string;
-  quorum: string;
+  // null when the subgraph has no quorum yet (e.g. a proposal that is not
+  // finalized) and the on-chain backfill in `useProposals` could not resolve it
+  quorum: string | null;
   transactionHash: string;
   voteCasts: {
     id: string;
--- a/apps/govern/components/Proposals/utils.ts
+++ b/apps/govern/components/Proposals/utils.ts
@@ -24,9 +24,9 @@ export const VOTES_SORTED: VoteSupport[] = [
   VoteSupport.Abstain,
 ];
 
-export const formatWeiToEth = (value: string) =>
+export const formatWeiToEth = (value: string | null | undefined) =>
   formatWeiNumber({
-    value: ethers.formatUnits(value, 18),
+    value: value == null ? undefined : ethers.formatUnits(value, 18),
     maximumFractionDigits: 3,
   });
 
@@ -56,6 +56,11 @@ export const getUserVote = (proposal: Proposal, address: Address) =>
  */
 export const isQuorumReached = (proposal: Proposal) => {
   const { votesFor, quorum } = proposal;
+  // Quorum can be unavailable (e.g. a not-yet-finalized proposal whose on-chain
+  // quorum could not be backfilled). Without it we cannot confirm the quorum was
+  // reached, so treat it as not reached rather than crashing on BigInt(null).
+  if (quorum == null) return false;
+
   const votesForBigInt = BigInt(votesFor);
   const quorumBigInt = BigInt(quorum);
 
--- a/apps/govern/hooks/useProposals.ts
+++ b/apps/govern/hooks/useProposals.ts
@@ -13,7 +13,10 @@ import { getProposals } from 'common-util/graphql/queries';
 import { useAppSelector } from 'store/index';
 
 export const useProposals = () => {
-  const { data: block, isLoading: isBlockLoading } = useBlock();
+  // Proposals' startBlock/endBlock and GovernorOLAS live on mainnet, so the
+  // current block used for status/quorum comparisons must come from mainnet too
+  // (otherwise a non-mainnet wallet yields meaningless block-number comparisons).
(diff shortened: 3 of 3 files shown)


kupermind (approved):
### Review — looks good ✅

Traced the full render path; the fix is correct and minimal.

- `formatWeiToEth(null)` → `formatWeiNumber({ value: undefined })` → `isNil(value)` short-circuits to `'0'`. No throw, Quorum column shows `0`.
- `isQuorumReached` returning `false` on `quorum == null` lands the legacy proposal on the `Defeated` tag (it isn't executed/cancelled/queued/ongoing), which is the correct status.
- Pinning `useBlock({ chainId: mainnet.id })` makes the `block.number` vs `startBlock`/`endBlock` comparisons apples-to-apples regardless of the connected wallet chain — fixes the stale caveat that was previously documented.
- The `|| null` normalization is sound: an item only enters `quorumArgs` when `!item.quorum`, so the early `return proposals.proposalCreateds` branch can never carry a null quorum.

**Nit (non-blocking):** `apps/govern/CLAUDE.md`'s Quorum-column note still says *"`useBlock()` there is not pinned to `mainnet.id`, so status/approximation can be wrong if the wallet is on a non-mainnet chain."* — that's exactly what this PR fixes, so the note is now stale and worth updating per the repo's "update the app's CLAUDE.md" convention.

atepem:
## Code review

**What it does:** Fixes the white-screen crash on the Govern proposals page caused by a legacy Defeated proposal with no quorum value (`BigInt(null)` / `ethers.formatUnits(null)` threw). It types `quorum` as `string | null`, makes `isQuorumReached` and `formatWeiToEth` null-safe, normalizes the backfill fallback to `null`, and pins `useBlock` to `mainnet.id`.

The fix is sound and well-scoped. I verified that `formatWeiNumber` returns `'0'` for `undefined` (via `isNil`), all `formatWeiToEth` callers pass values that satisfy the widened type, and the un-normalized return path (`useProposals.ts:72`) is only reachable when every proposal already has a truthy quorum, so no `null` leaks there.

### Findings

**1. `apps/govern/CLAUDE.md` (Notes section) — documentation now contradicts the code.**
The govern `CLAUDE.md` "Notes" explicitly states: *"`useBlock()` there is not pinned to `mainnet.id`, so status/approximation can be wrong if the wallet is on a non-mainnet chain."* This PR pins `useBlock({ chainId: mainnet.id })`, making that note false. The root `CLAUDE.md` requires: *"When adding a … hook … or other significant change to an app, update that app's `CLAUDE.md` to reflect the change."* Please update that note to reflect that the block is now pinned to mainnet.

**2. `components/Proposals/utils.ts:62` — null quorum from a *transient* backfill failure renders a passed proposal as "Defeated" (low severity, acknowledged tradeoff).**
`isQuorumReached` returns `false` whenever `quorum == null`. For the legacy proposal this is intended. But the backfill in `useProposals.ts` only populates `quorumByProposalId` when `quorumRes.result` is truthy — a transient RPC failure on the `quorum()` multicall leaves it `null`, so a finished, quorum-passing, not-yet-queued proposal would show **Defeated** instead of "Waiting to queue" until refetch. This is strictly better than the prior crash and the PR explicitly accepts the legacy case, so it's worth noting but not blocking.

### Verdict
The fix is correct. The only actionable item is updating the stale `apps/govern/CLAUDE.md` note (finding 1), ideally in this PR since it documents the exact behavior being changed.

atepem (approved):

PR #437, opened 2026-07-05
bond/operate/build: add missing not-legal pages

DavidMinarsch:
The shared middleware redirects restricted regions to /not-legal, but none of bond/operate/build actually had that page — redirected visitors landed on a 404 (verified live: all three return 404 on /not-legal). This adds the page to each app, following the marketplace app's existing pattern, with generalized wording (the route now serves both the sanctions list and the GB restriction).

Side effect that matters: touching apps/bond, apps/operate and apps/build forces Vercel to redeploy all three — #436 changed only libs/, and bond's deploy appears to have been skipped by the ignored-build-step, which would explain why the GB redirect works on operate/build but not on bond.

### DIFF
--- a/apps/bond/pages/not-legal.jsx
+++ b/apps/bond/pages/not-legal.jsx
@@ -11,7 +11,18 @@ const NotLegal = () => (
     <Result
       status="warning"
       title="Access restricted"
-      subTitle="Due to applicable legal and regulatory requirements, access to this site is not available from your current region."
+      subTitle={
+        <>
+          Due to applicable legal and regulatory requirements, access to this site is not available
+          from your current region. We use your approximate location data (region-level IP
+          information) solely for the purpose of enforcing this restriction and do not store this
+          information. For more details, please refer to our{' '}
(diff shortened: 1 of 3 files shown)


apps/bond/pages/not-legal.jsx
```diff
@@ -0,0 +1,19 @@
...
+  position: relative;
+  top: 100px;
+`;
+
+const NotLegal = () => (
+  <NotLegalContainer>
+    <Result
+      status="warning"
+      title="Access restricted"
+      subTitle="Due to applicable legal and regulatory requirements, access to this site is not available from your current region."
```
DavidMinarsch: on olas.network we also say "We use your approximate location data (region-level IP information) solely for the purpose of enforcing this restriction and do not store this information. For more details, please refer to our [Privacy Policy](https://olas.network/disclaimer)."

DavidMinarsch: Added to all three pages: the same region-level IP wording as olas.network/restricted, with the Privacy Policy link pointing at olas.network/disclaimer.

PR #441, opened 2026-07-31
fix(marketplace): [BDMARK-2240] source service counters from both-paths subgraph fields + clear audit advisories

atepem:
Two related changes: the marketplace counter fix, plus the supply-chain advisories that were blocking its CI. Combined into one PR at review request.

---

# 1. Marketplace service counters

## Why

Mech requests are moving off-chain. After the switch the marketplace subgraph stops creating per-request `Request` entities, and `Service.totalRequests` / `Service.totalDeliveries` **freeze** — they are incremented only by the on-chain `handleMarketplaceRequest` and the legacy `handleMarketplaceDelivery` handlers, never by `handleMarketplaceDeliveryWithSignatures`.

Two consequences:

1. The **Marketplace Role** tag stops updating — a service processing hundreds of off-chain requests shows the same numbers forever.
2. The **ERC8004 route gates the A2A card and MCP descriptor** on `totalDeliveries >= 1`, so an off-chain-only service silently loses both public endpoints.

This is `docs/consumer_migration.md` §9.2 in mech-analytics ("autonolas-frontend-mono — OLAS marketplace UI"), which documents the break but has no §11 work breakdown.

## What changed

`common-util/graphql/services.ts` now reads the counters that **both** handlers bump:

| Side | Source | Note |
|---|---|---|
| Demand | `Sender.totalLegacyRequests`, summed over `latestMultisig` + `historicalMultisigs` | `totalMarketplaceRequests` deliberately **not** added — the on-chain handler bumps both at once, so the sum double-counts every on-chain request |
| Supply | `Mech.totalDeliveriesTransactions` | `Mech.id` **is** the service id, so it joins directly on the requested ids |

Both merged with the legacy `Service.*` totals via `Math.max` so services whose history only exists under the old counters keep their role.

The `Service` type keeps its shape, so **`getMarketplaceRole()` and all three ERC8004 routes are unchanged** — the gate unfreezes with the source swap. No subgraph change required.

## Verification against the live gnosis subgraph

**Demand is exact parity:**

| Service | `Service.totalRequests` | `Sender.totalLegacyRequests` |
|---|---|---|
| 10 | 3796 | 3796 |
| 1000 | 13496 | 13496 |
| 1001 | 13104 | 13104 |

`totalMarketplaceRequests` equalled `totalLegacyRequests` on every sender sampled — confirming the double-count trap empirically.

**Supply is not parity, which is what makes the `Math.max` load-bearing rather than defensive:**

- Service 1722 — `Service.totalDeliveries` 619 vs `Mech.totalDeliveriesTransactions` 38
- Services 1698 / 1812 — 4 deliveries each, and **no `Mech` entity at all** (legacy agent-mech services)

Without the max, 1698 and 1812 would lose the Supply role outright.

**`totalOffChainRequests` is still 0 across sampled senders**, so this is numerically a no-op today — it only starts to matter after the switch.

## Not in scope

- **The activity table's Request/Delivery Data columns.** Those are IPFS links only (`AddressLink … isIpfs` builds an `href`, never fetches). Off-chain payloads stay private by design, so those cells render `NA` — correct behaviour, and the content must not be re-sourced (mech-analytics PR #18 closed on exactly that privacy ruling).
- **A mech-analytics integration.** It cannot serve this app: non-predict rows never reach `per_request_scores`, and its own counters are read *from* this same subgraph.

---

# 2. Supply-chain advisories

Two HIGH advisories were failing `yarn audit (production deps)` on `main`, blocking this PR and every other one.

**postcss** (GHSA-r28c-9q8g-f849, path traversal via `sourceMappingURL` auto-loading) — bumped the **existing** resolution `8.5.12` → `8.5.25`. Same minor line, no API change.

**brace-expansion** (GHSA-mh99-v99m-4gvg, unbounded expansion → OOM) — allowlisted, after trying the resolution first:

- The fix exists **only on the 5.x line** (`>=5.0.8`). `maintenance-v2` tops out at `2.1.4`, still inside the advisory range — no in-major upgrade path.
- Resolving to `5.0.9` **breaks the build**: `NX  expand is not a function`. brace-expansion 5.x's CJS export is not a drop-in for the 2.x default export that `minimatch` 5.x calls.

Allowlisted with review date 2026-10-31, on the same reachability argument as the two existing `js-yaml` entries. All 10 paths sit behind `minimatch` in build-time tooling (nx, ejs/jake, sucrase, expo-constants) operating on our own config globs — never untrusted input, and none in the shipped Next.js runtime bundle.

---

## Test plan

- [x] `npx tsc --noEmit` clean
- [x] `nx lint marketplace` — 0 errors (14 pre-existing warnings, none in changed files)
- [x] `nx run marketplace:build` — succeeds on postcss 8.5.25 (this build is what caught the brace-expansion break)
- [x] Queries executed against the live gnosis subgraph (see above)
- [x] `yarn audit:prod` — OK (3 allowlisted, no unlisted high/critical)
- [x] `yarn lint:lockfile` — no issues
- [x] `yarn audit:install-hooks` — OK (8 allowlisted); regenerated allowlist byte-identical, no new install hooks
- [x] `nx test marketplace` — 8 suites / 26 tests fail, but **identical on clean `main`** (verified by stash + rerun). Pre-existing, unrelated.
- [ ] **No unit tests added** for `services.ts`. There are none today. The mapping logic (max merge, multisig dedup/lowercasing, sum across historical multisigs, missing-`Mech` → 0) is worth covering with a mocked client — happy to add here or as a follow-up.


apps/marketplace/common-util/graphql/services.ts
```diff
@@ -32,13 +56,55 @@ export const getQueryForServiceDetails = ({ serviceIds }: { serviceIds: string[]
...
+        totalDeliveriesTransactions
+      }
     }
 `;
 
-type GraphQLResponse<T> = {
-  services: T[];
+export const getQueryForSenderCounters = ({ multisigs }: { multisigs: string[] }) => `
+    {
+      senders(
```
Tanya-atatakai, line 80: can we add `first: 1000` here? without it the-graph silently caps at 100 rows — `serviceIds` is an unbounded client-supplied query param (`pages/api/services.ts:20-23`), so 100+ multisigs is reachable, and a truncated `senders` page doesn't error, it just undercounts. today the `Math.max` hides it, but post-switch that's exactly the silent role-loss this PR exists to prevent.

```suggestion
      senders(
        first: 1000
```

apps/marketplace/common-util/graphql/services.ts
```diff
@@ -32,13 +56,55 @@ export const getQueryForServiceDetails = ({ serviceIds }: { serviceIds: string[]
           address
         }
       }
+      meches(
```
Tanya-atatakai, line 66: same here for `meches`

```suggestion
      meches(
        first: 1000
```

Tanya-atatakai (commented):
Verified the counter logic independently before commenting — good catch on the double-count trap, and the `Math.max` reasoning is solid: `incrementSenderRequests` in the subgraph's `utils.ts:1192-1199` bumps `totalLegacyRequests` **and** `totalMarketplaceRequests` together on-chain, while the signed-delivery handler (`mech-marketplace.ts:283-284`) bumps `totalLegacyRequests` + `totalOffChainRequests` only — so legacy-only is the right demand field and the sum would double-count, exactly as described. Re-ran the sampled queries against the live gnosis subgraph: all numbers reproduce (3796/13496/13104 parity, 1722 = 619 vs 38, 1698/1812 have no `Mech` entity, `totalOffChainRequests` = 0). The brace-expansion allowlist reasoning (5.x-only fix, tested break) is convincing and matches the js-yaml precedent.

**Verdict: approve-with-caveat** — please add the `first` caps (inline comments); the rest is optional.

### Outside this PR

- the pre-existing `services(...)` query has the same missing-`first` 100-cap (untouched behaviour) — worth a ticket together with the new ones if only the new queries are fixed here
- **cross-repo, important:** olas-website's `roi-distribution.ts:349` computes `senderTotal = totalLegacyRequests + totalMarketplaceRequests` against this same subgraph. Per the handler code above both fields bump together on every on-chain request — and on the live subgraph they're exactly equal on every trader safe I sampled — so that sum is ~2x the real count, which inflates `settledMechRequests`. Needs its own check/fix in olas-website
- string-interpolated GraphQL (`toQuotedList`) rather than variables is the file's pre-existing pattern — fine here, could be a cleanup ticket

### What I verified

- fetched the marketplace subgraph mapping source (`mech-marketplace.ts`, `utils.ts`) from `valory-xyz/autonolas-subgraph` and read the actual counter bumps for both handlers
- ran the PR's verification queries against the live gnosis marketplace subgraph — reproduced every sampled number, confirmed `meches` is the real collection name and `latestMultisig` values arrive lowercase (so the `.toLowerCase()` join is safe)
- checked all four callers of `getServicesFromMarketplaceSubgraph`; `totalRequests`/`totalDeliveries` feed only `getMarketplaceRole()`'s `> 0` checks (`ListServices/utils.tsx:95-107`) — nothing renders them as numbers, so the "gate signal, not a count" claim holds
- confirmed `serviceIds` is an unbounded query param at `pages/api/services.ts:20`

apps/marketplace/CLAUDE.md
```diff
@@ -103,6 +103,31 @@ Tracks **requests** (demand) and **deliveries** (supply) for mech services.
...
+`common-util/graphql/services.ts` therefore reads the counters that are incremented on **both** paths, and merges them with the legacy `Service.*` totals via `Math.max` so services with only pre-switch history keep their role:
+
+| Role side | Source | Notes |
+|---|---|---|
+| Demand | `Sender.totalLegacyRequests`, summed over the service's `latestMultisig` + `historicalMultisigs` | Do **not** add `totalMarketplaceRequests` on top — the on-chain handler bumps both at once, so the sum over-counts every on-chain request |
+| Supply | `Mech.totalDeliveriesTransactions` | Incremented by delivered item count on the delivery mech by both handlers. `Mech.id` **is the service id**, so it joins directly on `serviceIds` (the address lives in `Mech.address`) |
+
+This needs two round trips: the multisigs are only known once `services` returns. `getMarketplaceRole()` and the ERC8004 routes are unchanged — the `Service` type keeps its `totalRequests` / `totalDeliveries` shape, only the source changed.
+
+Verified against the live gnosis subgraph:
```
Tanya-atatakai: tiny, up to you — this verification block with live numbers ("619 vs 38", "`totalOffChainRequests` is still 0") is a dated snapshot that will quietly rot in an undated doc. the numbers already live in the PR description; maybe keep only the doctrine here (which fields, the double-count trap, "max is a gate signal not a count") and drop the sampled values? wdyt

package.json
```diff
@@ -178,7 +178,7 @@
     "micromatch": "4.0.8",
     "min-document": "2.19.1",
     "nanoid": "3.3.11",
-    "postcss": "8.5.12",
+    "postcss": "8.5.25",
```
Tanya-atatakai, line 181: the postcss bump makes it declare `nanoid: ^3.3.16`, but our resolutions pin nanoid to `3.3.11` — the lockfile now has `nanoid@^3.3.16: version "3.3.11"`, which violates the declared range. works today b/c yarn resolutions override semver, but can we bump the resolution to `3.3.16` so we're not pinning below a dep's floor?

.supply-chain/audit-allowlist.json
```diff
@@ -27,6 +27,15 @@
...
       "added": "2026-07-23",
       "review": "2026-10-23"
+    },
+    {
+      "id": 1124334,
+      "ghsa": "GHSA-mh99-v99m-4gvg",
+      "package": "brace-expansion",
+      "severity": "high",
+      "reason": "DoS (unbounded expansion length -> OOM crash) on crafted glob patterns. All 10 paths reach it through minimatch in build-time tooling only (@nx/next > @nx/js > minimatch; @nx/devkit > ejs > jake > minimatch; tailwindcss > sucrase > glob > minimatch; expo-constants > @expo/fingerprint > minimatch) — the patterns are our own config globs, never untrusted input, and none of this is in the shipped Next.js runtime bundle. The fix is only on the 5.x line (>=5.0.8); the maintenance-v2 line tops out at 2.1.4, still inside the advisory range. Resolving to 5.0.9 was tested and breaks the build with 'expand is not a function' — brace-expansion 5.x's CJS export is not a drop-in for the 2.x default export that minimatch 5.x calls. Revisit when minimatch ships a release depending on brace-expansion >=5.0.8.",
+      "added": "2026-07-31",
```
Tanya-atatakai: tiny, `"added": "2026-07-31"` was in the future when this was pushed :D

atepem:
Thanks — all five addressed. Two of them turned into bigger changes than expected.

### Page caps (required) — done

Added `first: 1000` to `senders` and `meches` via a shared `PAGE_LIMIT`. Agreed on the reasoning: a truncated page doesn't error, and post-switch the `Math.max` stops hiding it.

I also applied it to the pre-existing `services(...)` query rather than filing a ticket. Same one-line bug class, same unbounded `serviceIds` input, same query builder — leaving a known silent-truncation path in the function I was already rewriting seemed worse than the scope creep. Shout if you'd rather I back that one out.

### nanoid — good catch, done

Confirmed: postcss 8.5.25 declares `nanoid: ^3.3.16`, and the lockfile carried `nanoid@^3.3.16: version "3.3.11"`. Resolution bumped to `3.3.16`.

### brace-expansion — the advisory was re-issued, so it's a real fix now

Chasing the nanoid bump made the audit fail with a *different* advisory ID. GHSA-mh99-v99m-4gvg has been re-published with a corrected range:

| | old (`1124334`) | new (`1130589`) |
|---|---|---|
| affected | `<=5.0.7` | `>=2.0.0 <2.1.3` |
| fix in | `>=5.0.8` | **`>=2.1.3`** |

Upstream backported to the 2.x line, so the 5.x jump that broke the build isn't needed. Resolved `2.1.2` → `2.1.4` and **deleted the allowlist entry entirely** — the audit is back to the two original js-yaml entries. That removes the weakest part of this PR, so thanks for the prod that surfaced it.

### CLAUDE.md snapshot — agreed, done

You're right that it would rot. Kept the doctrine (which fields, the double-count trap, "gate signal not a count") and dropped the sampled figures, with a pointer here for the numbers.

### The date

Minor, but the entry wasn't future-dated — the commit is authored `2026-07-31 14:53:27 +0100` and it said `2026-07-31`. Moot either way now that the entry is gone.

---

### On the olas-website finding — this one matters

Your `roi-distribution.ts:349` catch looks right and I don't think it should ride on this PR. `senderTotal = totalLegacyRequests + totalMarketplaceRequests` is the exact double-count this PR avoids, and it feeds `settledMechRequests`, so it's a live correctness bug in a shipped metric rather than a post-switch risk. Worth its own issue against olas-website — happy to open it with the handler citations if you haven't already.

### Re-verified after the changes

`yarn audit:prod` OK (2 allowlisted) · `lint:lockfile` clean · `audit:install-hooks` OK (8, no new hooks) · `tsc` clean · lint 0 errors · `nx run marketplace:build` succeeds · all three capped queries re-run against the live gnosis subgraph, identical values (1722 → 619 vs 38, sender → 3796).

Still open: no unit tests for `services.ts`. Say the word and I'll add them here rather than as a follow-up.

atepem:
### Review status — all 5 comments addressed (`7dd98d2d`)

| Comment | Status |
|---|---|
| `first: 1000` on `senders` | ✅ added |
| `first: 1000` on `meches` | ✅ added |
| nanoid pinned below postcss's `^3.3.16` floor | ✅ resolution → `3.3.16` |
| CLAUDE.md dated snapshot will rot | ✅ sampled figures dropped, doctrine kept |
| `"added"` date in the future | ➖ moot — entry deleted (see below); commit was authored same-day `2026-07-31 14:53 +0100` |

**Two things went further than the comments asked:**

- Also capped the **pre-existing** `services(...)` query — same bug class, same unbounded `serviceIds` input. Extracted as a shared `PAGE_LIMIT`. Easy to back out if you'd rather keep the diff tight.
- **brace-expansion is now a real fix, not an allowlist.** GHSA-mh99-v99m-4gvg was re-issued while this PR was open (`1124334` → `1130589`), range corrected to `>=2.0.0 <2.1.3`, fixed in `>=2.1.3`. Upstream backported to 2.x, so the 5.x jump that broke the build is unnecessary — resolved `2.1.2` → `2.1.4` and **deleted the allowlist entry**. Audit is back to the two original js-yaml entries.

**CI: all checks passing.** Re-verified locally too — `tsc` clean, lint 0 errors, `marketplace:build` succeeds, and the three capped queries re-run against the live gnosis subgraph return identical values.

Still open, unchanged: no unit tests for `services.ts`. And the olas-website `senderTotal` double-count you spotted needs its own issue — happy to open it.

apps/marketplace/common-util/graphql/services.ts
```diff
@@ -49,11 +124,43 @@ export const getServicesFromMarketplaceSubgraph = async ({
...
+  );
+
+  // The multisigs are only known once the services come back, so the sender
+  // counters need a second round trip.
+  const multisigs = Array.from(
+    new Set((response.services ?? []).flatMap((service) => getServiceMultisigs(service))),
+  );
+
+  const requestsByMultisig = new Map<string, number>();
+  if (multisigs.length > 0) {
```
DIvyaNautiyal07, line 140: P2 — `PAGE_LIMIT` caps the `senders` query but doesn't guarantee it never truncates. `services` and `meches` are safe because they're bounded by `serviceIds` (URL-length bounded in practice), but `multisigs` is `union(latestMultisig, historicalMultisigs across all services)` and isn't bounded by input. A service that has rotated multisig many times, multiplied across many services, can exceed 1000 — the second query then silently drops senders, `requestsByMultisig` under-counts, and `Math.max` masks it with the legacy count. Same silent-truncation failure class this PR exists to prevent.

A one-line `console.warn` when `multisigs.length >= PAGE_LIMIT` makes the failure detectable. The proper long-term fix (capping `parsedServiceIds.length` at `pages/api/services.ts:20-23`) would also close the deferred injection concern in one place.

apps/marketplace/common-util/graphql/services.ts
```diff
@@ -49,11 +124,43 @@ export const getServicesFromMarketplaceSubgraph = async ({
...
+    const senderResponse = await client.request<SenderCountersResponse>(
+      getQueryForSenderCounters({ multisigs }),
+    );
+    for (const sender of senderResponse.senders ?? []) {
+      requestsByMultisig.set(sender.id.toLowerCase(), toCount(sender.totalLegacyRequests));
+    }
+  }
+
+  return (response.services ?? []).map((service) => {
+    const requestsFromSenders = getServiceMultisigs(service).reduce(
```
DIvyaNautiyal07, line 161: P2 — the `totalMarketplaceRequests` exclusion is the single most important invariant in this PR, and no test would fail if someone added `+ sender.totalMarketplaceRequests` to the reduce. Once off-chain traffic starts flowing, the regression silently doubles the count.

One unit test — mock a sender response with both `totalLegacyRequests: "100"` and `totalMarketplaceRequests: "500"`, assert `totalRequests === 100` — locks the trap in. You offered to add tests here; this is the highest-leverage single case if you only add one.

DIvyaNautiyal07 (dismissed):

Tanya-atatakai (dismissed):

atepem:
Both P2s addressed in `35a80e24` — thanks @DIvyaNautiyal07, the second one was the right call.

**Multisig truncation.** Agreed the cap isn't a guarantee: `multisigs` is the union of every service's multisig history and isn't bounded by input the way `services`/`meches` are. Added the `console.warn` at `multisigs.length >= PAGE_LIMIT` so it's detectable rather than silent. Left the `parsedServiceIds` cap in `pages/api/services.ts` as a follow-up, per your note that it closes the injection concern in the same place — happy to do it here instead if you'd rather it not linger.

**The double-count test.** Added, and **verified by mutation** rather than just asserting it passes: re-introducing `+ totalMarketplaceRequests` into the reduce fails that one test (`Expected: 100`, `Received: 600`) and no other. So the trap is genuinely locked, not just documented.

That test came with 10 more — first coverage this module has had:

- multisig dedupe across `latest` + `historical` (same safe listed twice counts once)
- case-insensitive sender matching
- legacy-total fallback when it's higher
- no-multisig short circuit — asserts the second query is *skipped*, not just that the result is right
- Mech-ahead-of-Service, and the missing-`Mech` case that makes `Math.max` load-bearing
- cross-service Mech attribution (service 1 must not pick up service 2's row)
- metadata / mechAddresses defaults

Full suite: **26 failed / 56 passed**, against a **26 failed / 45 passed** baseline on `main` — same pre-existing failures, +11 passing. `tsc` clean, lint 0 errors.

Tanya-atatakai (approved):

ROMAN-VALORY (approved):

PR #442, opened 2026-07-31
fix: [BDMARK-2237] clear brace-expansion and postcss high-severity advisories

atepem:
## Why

Two new HIGH advisories are failing the `yarn audit (production deps)` gate on `main`, blocking every open PR.

## postcss — fixed by bump

**GHSA-r28c-9q8g-f849** — path traversal in previous-source-map auto-loading (`sourceMappingURL`) leading to arbitrary `.map` file disclosure.

Bumped the **existing** resolution `8.5.12` → `8.5.25`. Same minor line, no API change. Reached via `next > postcss` and `@binance/w3w-* > tailwindcss > postcss`.

## brace-expansion — allowlisted, with a tested reason

**GHSA-mh99-v99m-4gvg** — DoS via unbounded expansion length causing an OOM crash.

I tried the resolution first. It does not work:

- The fix exists **only on the 5.x line** (`>=5.0.8`). The `maintenance-v2` line tops out at `2.1.4`, still inside the advisory range — so there is no in-major upgrade path.
- Resolving to `5.0.9` **breaks the build**: `NX  expand is not a function`. brace-expansion 5.x's CJS export is not a drop-in for the 2.x default export that `minimatch` 5.x calls.

So it is allowlisted with a review date of 2026-10-31, on the same reachability argument as the two existing `js-yaml` entries. All 10 paths sit behind `minimatch` in build-time tooling:

```
@nx/next > @nx/js > minimatch > brace-expansion
@nx/next > @nx/devkit > ejs > jake > minimatch > brace-expansion
@binance/... > tailwindcss > sucrase > glob > minimatch > brace-expansion
@solana/... > expo-constants > @expo/fingerprint > minimatch > brace-expansion
```

Those globs are our own config patterns, never untrusted input, and none of this is in the shipped Next.js runtime bundle. Revisit when `minimatch` ships a release depending on `brace-expansion >=5.0.8`.

## Verification

- [x] `yarn audit:prod` — OK (3 allowlisted, no unlisted high/critical)
- [x] `yarn lint:lockfile` — no issues
- [x] `yarn audit:install-hooks` — OK (8 allowlisted); regenerated allowlist is byte-identical, so no new install hooks
- [x] `nx run marketplace:build` — succeeds on postcss 8.5.25 (this is what caught the brace-expansion break)

CI will exercise the remaining app builds.

## Note

This unblocks #441, which is currently red only on this gate.

### DIFF
--- a/.supply-chain/audit-allowlist.json
+++ b/.supply-chain/audit-allowlist.json
@@ -27,6 +27,15 @@
       "reason": "Same GHSA on the js-yaml 3.x line, reached via nx build-time tooling (@nx/next > nx > @yarnpkg/parsers > js-yaml 3.x). Build-time only, trusted input, not in the shipped runtime. Cannot be co-resolved with the 4.x consumer via a single Yarn resolution without breaking API compatibility.",
       "added": "2026-07-23",
       "review": "2026-10-23"
+    },
+    {
+      "id": 1124334,
+      "ghsa": "GHSA-mh99-v99m-4gvg",
+      "package": "brace-expansion",
+      "severity": "high",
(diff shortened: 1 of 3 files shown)


atepem:
Folded into #441 at review request — the marketplace branch is now rebased on top of this commit, so both changes land as a single PR. Commit `c9009bae` is unchanged and carried over verbatim.

